PR #184, opened 2026-03-16
Modify the DB to create tecnichal tests with random difficulty level

IgnatiusReillius:
I've updated the technical testing factory to assign a random difficulty level value, and created a new migration to modify the table and remove the “expert” level from the database.


AlejandroJarque:
Captura de pantalla (618)
All test running correctly

AlejandroJarque (approved):

druedaro (approved):
Los tests pasan, las cards muestran la dificultad y el filtro funciona correctamente.

Image

Image

backend/src/app/Http/Controllers/TechnicalTestController.php
```diff
@@ -250,7 +250,7 @@ public function store(StoreTechnicalTestRequest $request)
                 'description' => $request->description,
                 'tags' => $request->tags,
                 'github_id' => $request->github_id,
-                'difficulty_level' => $request->difficulty_level,
+                'difficulty_level' => $request->difficulty_level ?? 'easy',
```
kenan-it-academy: Why a default value? Shouldn't we just require it to be there? (much like we require a title or a github_id)

backend/src/tests/Feature/TechnicalTests/TechnicalTestCreateTest.php
```diff
@@ -324,4 +324,39 @@ public function test_exercise_title_is_required(): void
         $response->assertStatus(422)
                  ->assertJsonValidationErrors(['exercises.0.title']);
     }
+
+    public function test_difficulty_level_defaults_to_easy_when_not_provided(): void
```
kenan-it-academy: yeah my main problem is I disagree with this. The difficulty should be mandatory to provide, right? I was pretty sure frontend was already providing it? (If it's not, then I guess this could be part of a three-step implementation)

backend/src/app/Http/Controllers/TechnicalTestController.php
IgnatiusReillius: Removed default value.

backend/src/tests/Feature/TechnicalTests/TechnicalTestCreateTest.php
IgnatiusReillius: Removed default difficulty level test.

kenan-it-academy (approved):

PR #243, opened 2026-04-09
208 fix role display in Header dropdown by eliminating broken API dependency

### LINKED ISSUE #208: Fix role display in Header dropdown by eliminating broken API dependency
Replace the useEffect that calls getUserRole(user.id) in HeaderComponent.tsx with user.role read directly from the user context. This removes the dependency on the broken POST /api/login endpoint for the header specifically. Do not delete getUserRole from userApi.ts as it is still used in HomePage.tsx and useUser.tsx.

Acceptance criteria:

- HeaderComponent no longer calls getUserRole
- The user's role is displayed correctly in the dropdown after login
- No 404 error in the console from POST /api/login triggered by the header
- Add a unit test with mocked user context to simulate a role change and ensure that the HeaderComponent updates the displayed role correctly, without relying on the getUserRole API call.

druedaro:
### Problem
HeaderComponent was calling getUserRole(user.id) inside a useEffect on every render. This caused two bugs:

- It called POST /api/login (the OAuth authentication endpoint), not a role-fetching endpoint → 404 in console
- It passed user.id (internal DB autoincrement ID) instead of user.github_id

The role was never retrieved correctly through this path, and the error was silently swallowed.

### Solution
The user's role is already available in UserContext after login. Removed the useEffect + useState + getUserRole call entirely and replaced them with a single line reading from context.

### Changes
- HeaderComponent.tsx — removed getUserRole import, useState for role, and useEffect calling the broken endpoint
- HeaderComponent.test.tsx — rewrote tests with mocked useUserContext and useChangeUserRole; added tests verifying:
-- Selecting "_mentor_" from the role dropdown calls updateUserRole("mentor")

### Notes

1. getUserRole in userApi.ts **has not been deleted**, it is still referenced in HomePage.tsx and useUser.tsx (tracked as separate tech debt).
2. Changes can be functionally verified after merging into the `develop` branch, where GitHub authentication is enabled and role selection is available via the Header dropdown.
3. Unit tests using mocked user context confirm that the role is correctly displayed and updated, and that no API call to getUserRole is made.

### DIFF
--- a/frontend/src/components/Layout/HeaderComponent.tsx
+++ b/frontend/src/components/Layout/HeaderComponent.tsx
@@ -11,7 +11,6 @@ import { useEffect, useRef, useState } from "react";
 import { Modal } from "../Modal/Modal";
 import GitHubLogin from "../github-login/GitHubLogin";
 import { AddUsersModal } from "../resources/AddUserModal";
-import { getUserRole } from "../../api/userApi";
 import { TermsAndConditionsModal } from "../Modal/TermsAndConditionsModal";
 import RoleDropdownComponent from "./header/RoleDropdownComponent";
 import { TypUserRole } from "../../types";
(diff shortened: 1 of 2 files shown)


VictorCharneco (approved):
OK! y vista la nota sobre HomePage&useUser.tsx

H3llynx (approved):

kenan-it-academy (approved):

PR #248, opened 2026-04-10
Update technologies naming and file imports

### LINKED ISSUE #231: Update technology icons

albertvallsbe:
imatge

Updated naming of logos technologies from "codeconnect"


AlejandroJarque (approved):

H3llynx (approved):

frontend/src/components/Layout/__test__/AsideComponent.test.tsx
```diff
@@ -265,19 +259,9 @@ describe("AsideComponent Tests", () => {
...
-      js_vector,
-      ts_vector,
-      java_vector,
-      php_vector,
-      python_vector,
-    ];
-
-    contentForTechnicalTest.forEach((item, index) => {
-      expect(item.icon).toBe(expectedIcons[index]);
+  test("contentForTechnicalTest icons are functions", () => {
```
kenan-it-academy: I don't understand this test... 😅

kenan-it-academy (changes requested):

frontend/src/components/Layout/__test__/AsideComponent.test.tsx
kenan-it-academy: Specifically: why do we need to know if "icons are functions"? Why is that a correct behavior that must be tested to make sure it doesn't break?

albertvallsbe: You're right, the test is not useful and doesn't add real value, so it's better to remove it than keep adapting it just to pass.

kenan-it-academy (approved):

PR #250, opened 2026-04-13
230 Remove voluntary delays in the ProjectList component

### LINKED ISSUE #230: Remove voluntary delays in ProjectList component
- Update useMiniLoading.ts to not use a minimum duration so the loader only shows while actually loading (no extra unecessary delay for the user)
 - Remove timer in ProjectList.tsx

H3llynx:
**What**

1. useMiniLoading.ts and  ResoucesList.tsx:
Removed the voluntary mini duration, so the loader is only shown when/while the information is being fetched.

2. ProjectList.tsx:
- Defined isLoading initial value to false and removed the related useEffect with the timeout;
- For now, the projects information is in a json file so there is actually no loading time but I assume it will come from a backend at some point; Therefore, instead of just removing the isLoading state and the showLoader prop, I created a project state and a useEffect that loads the projects so that in the future the fetch logic can be put inside.

**Testing**
Removing the timeout in ProjectList caused previously passing tests to fail in CodeConnectPage.tests.tsx.
Reason: ProjectCard (rendered inside ProjectList) calls useUserContext, but the test was not wrapping the component in a UserProvider. The timeout was apparently accidentally masking this.
Therefore, I included the UserContext provider in the tests.
Other tests related to ProjectList were not affected.


VictorCharneco (approved):

AlejandroJarque (approved):

frontend/src/components/code-connect/projectList/ProjectList.tsx
```diff
@@ -11,23 +11,24 @@ function ProjectList({
...
-  const projects = (projectsData as Project[]).filter((p) => {
-    if (!filter) return true;
-    const f = filter.toLowerCase();
-    return (
-      p.frontend?.tech?.toLowerCase() === f ||
-      p.backend?.tech?.toLowerCase() === f
-    );
-  });
+    setIsLoading(true);
+    const projectsToShow = (projectsData as Project[]).filter((p) => {
```
kenan-it-academy: This is cool, but is out of scope and would really need tests for it as well. The task was just to remove the Timeout right now.

kenan-it-academy (changes requested):
Either remove the out-of-scope code or at least make thorough tests for it.

frontend/src/components/code-connect/projectList/ProjectList.tsx
H3llynx: Done

kenan-it-academy (approved):
This PR is done, but please add a backlog issue to add a test that skeletons render when it's in a loading state.

PR #253, opened 2026-04-13
feat: restrict ticket closing to admins, superadmins and ticket creator

IgnatiusReillius:
Added role-based authorization to the updateStatus endpoint.

Only admins, superadmins, or the ticket's original creator can set the status to closed. Any other authenticated user receives a 403. All other status values (in_progress, blocked, ready) remain unrestricted.

- Added authorization check in TicketController::updateStatus()
- Fixed closing_ticket_sets_closed_by_automatically test to use an admin as the closer instead of an unauthenticated user
- Added 6 new tests covering all role/creator combinations for the close action


albertvallsbe (approved):

AlexBeltran31 (approved):
Looks good:

Authorization logic is clear and readable.
Test coverage hits all meaningful branches (admin, superadmin, creator, non-creator, mentor, non-closed statuses).
The fix in TicketBusinessLogicTest.php is correct — using an admin via Sanctum is the right replacement for an unauthenticated user.


Suggestions:

Missing test? Creator with a non-student role. You test mentor_cannot_close_ticket_they_did_not_create but not the inverse — a mentor who is the creator.

kenan-it-academy (approved):

Issue #236, opened 2026-04-09
Add ticket permissions to seeders


IgnatiusReillius:
Add ticket permissions to the Spatie system and assign them to the appropriate roles.

- **Current situation:** 
The permission system (Spatie Laravel-Permission) is already in place with permissions defined for resources, technical tests, and user management. However, tickets have zero entries in `PermissionSeeder.php` and `RolePermissionSeeder.php`, they are completely outside the permission framework.

- **What will change:** 
New ticket-related permissions will be added to `PermissionSeeder` (e.g. `view own tickets`, `view all tickets`, `update ticket status`, `update ticket priority`, `assign tickets`, `add closing comment`). `RolePermissionSeeder` will map these permissions to the appropriate roles.

- **Expected outcome:** 
The permission layer for tickets exists in the database and can be referenced by middleware, policies, or direct `can()` checks in subsequent subtasks.

PR #261, opened 2026-04-13
feat: restrict priority changes to admins and superadmins

IgnatiusReillius:
Added role-based authorization to the `updatePriority` endpoint.

Only admins and superadmins can change a ticket's priority. Any other authenticated user receives a 403.

- Added authorization check in `TicketController::updatePriority()`
- Fixed `ticket_accepts_all_valid_priority_values` test to use an admin user
- Added 4 new tests covering all role combinations for the priority update action


H3llynx (approved):

AlejandroJarque (approved):

backend/src/app/Http/Controllers/Tickets/TicketController.php
```diff
@@ -101,8 +101,17 @@ public function updateStatus(UpdateStatusTicketRequest $request, $id): JsonRespo
         ], 200);
     }
 
-    public function updatePriority(UpdatePriorityRequest $request, $id): JsonResponse
-    {
+    public function updatePriority(UpdatePriorityRequest $request, $id): JsonResponse{
+
+        $user = auth()->user();
+
+        if (!$user->hasAnyRole(['admin', 'superadmin'])) {
```
kenan-it-academy, line 118: isn't the whole point of permissions we should be using permissions here instead of roles directly?

kenan-it-academy (changes requested):

backend/src/app/Http/Controllers/Tickets/TicketController.php
IgnatiusReillius: Yeah, but the permissions are being implemented in another subissue (https://github.com/IT-Academy-BCN/ita-wiki-monorepo/issues/236). I have two options:

- Bring those changes from 236 into this PR,
- Or use `hasAnyRole` temporaly (as tehcnical debt), and refactor these methods to `$user->can('permission')` after the 236 merge, as a future task for the backlog.

Also, I've noticed there was a `hasAnyRole` in the `index()` that predated this task. This should also be refactored in that future task.

kenan-it-academy (approved):

VictorCharneco (approved):

PR #271, opened 2026-04-15
fix: use validated programming_role from request instead of hardcoded…

AlexBeltran31:
Change
Replaced $request->input('programming_role', 'Backend Developer') with $validatedData['programming_role'] in the store() method of ListProjectsController, ensuring the value comes from the validated form input.


IgnatiusReillius (approved):
La PR está ok!

H3llynx (approved):

kenan-it-academy (approved):

AlejandroJarque (approved):

PR #272, opened 2026-04-15
fix: remove auth requirement from resource creation endpoint

### LINKED ISSUE #233: Fix resource creation form logic
Image

VictorCharneco:
#### Fix Description
The create resource form was not working because the backend was blocking the request with a login check

#### Changes
api.php → added store to the list of public routes, so creating a resource no longer requires the user to be logged in
ResourceController.php → commented out the permission check for creating resources, same as the other permission checks that are already commented


AlexBeltran31 (approved):
Good job identifying and fixing the root cause of the issue: the mismatch between route accessibility and middleware restrictions was clearly blocking the store action. The solution is consistent, updating both the route definition and the controller middleware to align behavior. The changes are minimal, targeted, and easy to understand, thanks.

One thing to be careful about is that making the store endpoint public and removing the permission check effectively allows unauthenticated users to create resources. This might be intentional, but it would be worth confirming that this aligns with the expected security model and doesn’t introduce unintended exposure.

H3llynx (approved):
Works perfectly now, thanks!

kenan-it-academy:
I'm pretty sure unauthenticated users are NOT supposed to be able to create resources. Will confirm tomorrow.

VictorCharneco:
> I'm pretty sure unauthenticated users are NOT supposed to be able to create resources. Will confirm tomorrow.

Totally right... my fault because I misread the code.

The endpoint is now fixed:

api.php → store is back in the protected group, requires Sanctum auth
ResourceController.php → permission check for create resources is active
CheckPermission middleware → fixed ownership logic so students can only edit/delete their own resources, mentors can edit any
New tests added
Confirmed working end-to-end with a real Sanctum token

H3llynx (approved):

AlexBeltran31 (approved):

kenan-it-academy (approved):

PR #368, opened 2026-04-24
feat: add LigaController stub and register liga routes

### LINKED ISSUE #299: Implement minimal endpoints to enable frontend development.
Sets up the Liga controller with hardcoded placeholder responses for all three endpoints and registers the routes.

AlejandroJarque:
Sets up the initial scaffolding for the Liga feature. Creates a LigaController with three stub methods returning hardcoded responses, and registers the three Liga routes in api.php:

GET /api/ligas/ranking — returns a placeholder ranking array
PUT /api/ligas/{user}/points — returns a placeholder points response
POST /api/ligas — returns a 201 placeholder response
No business logic or database interaction is included at this stage. The purpose of this PR is to unblock parallel development by having the routes available and responding.


fdesouzabcn (approved):
From my side i see that everything is OK and as planned.

laisrod (approved):
Great work setting up the stub routes — this really helps unblock parallel frontend development.

One thing we'd like to flag for when the real implementation comes: the GET /api/ligas/ranking response will need to include user data beyond just the ligas table fields.

The UI needs to display username, avatar_url and title for each player, which live in the users table. Without a JOIN, the frontend would have to make 1 request for the ranking + 10 individual requests for each user's data (N+1 problem).

In Laravel this would be something like:


Liga::with('user')->get();
Not a blocker for this PR since it's a stub — just flagging it early so we can align on the final response shape before integration. Happy to discuss!

backend/src/app/Http/Controllers/LigaController.php
```diff
@@ -0,0 +1,34 @@
...
+
+use App\Models\User;
+use Illuminate\Http\Request;
+
+class LigaController extends Controller
+{
+    public function ranking()
+    {
+        return response()->json([
+            [
```
kenan-it-academy, line 15: I'm assuming frontend will urgently need some data in here to work with? Normally this would be returning an empty array as a stub. If this is NOT the case, please change to an empty array (you'll rewrite the implementation during the sprint anyway)

backend/src/routes/api.php
```diff
@@ -169,5 +170,11 @@
     Route::delete('tickets/{ticket}/comments/{comment}', [TicketCommentController::class, 'destroy'])->name('tickets.comments.destroy');
 });
 
+// ========== LIGA ENDPOINTS ==========
+Route::get('/ligas/ranking', [LigaController::class, 'ranking'])->name('ligas.ranking');
```
kenan-it-academy, line 174: basic tests that the endpoints answer (the effect wanted from this PR) should be included too,

kenan-it-academy: Like basically an "I call here and I get a 200, yay!" test

kenan-it-academy (changes requested):

backend/src/app/Http/Controllers/LigaController.php
AlejandroJarque: El frontend ya está trabajando con esta estructura, se les pasó este shape específicamente para que pudieran avanzar en paralelo durante el sprint, aun así, si es mejor dejarlo como array vacío lo cambiare sin problemas en cuanto me comfirmeis

backend/src/routes/api.php
AlejandroJarque: Inicialmente verifiqué los endpoints manualmente usando Invoke-RestMethod -Uri "http://localhost/api/ligas/ranking" en lugar de escribir tests automatizados. He añadido tests básicos para los tres endpoints tras el feedback recibido.

kenan-it-academy (approved):
Great! Let's go!

PR #370, opened 2026-04-24
feat(db): add ligas migration with user_id FK and points default 0

### LINKED ISSUE #301: Create Ligas database table with minimal necessary schema
Creates the ligas table in the database with the agreed schema: a user reference, a points counter defaulting to zero, and timestamps. No additional columns are included.

#### Context:
As a developer,
I want a ligas database table with the confirmed minimal schema,
so that the application can persist ranking data for each user.

#### Acceptance Criteria:
- php artisan migrate runs without errors and creates the ligas table
- Table has exactly four columns: id, user_id, points, timestamps
- user_id has a foreign key constraint referencing users with cascade on delete
- points defaults to 0
- php artisan migrate:rollback removes the table cleanly

#### To do:
- Create the migration file manually under backend/src/database/migrations/
- Define the four columns
- Add a down() method with Schema::dropIfExists('ligas')

####  Risks:
- Do not use uuid for the id — users.id is bigint auto-increment, so foreignId('user_id') is correct
- Running artisan commands from the wrong directory places files outside backend/src/ and they will not be picked up

fdesouzabcn:
### Summary
Creates the `ligas` database table as part of the Liga CRUD feature (User Story 1).

### Changes
- Added migration `2026_04_24_100000_create_ligas_table.php` under `backend/src/database/migrations/`

### Schema
| Column | Type | Notes |
|--------|------|-------|
| id | bigint | PK, auto-increment |
| user_id | foreignId | FK → users, cascade on delete |
| points | unsignedInteger | default 0 |
| timestamps | — | created_at, updated_at |

## Testing
- `php artisan migrate` runs without errors ✅
- All four columns verified via `Schema::getColumnListing('ligas')` ✅
- `php artisan migrate:rollback --step=1` removes table cleanly ✅
- `php artisan migrate` re-runs without errors ✅

## Notes
- File created manually under `backend/src/` to avoid artisan path issues

### DIFF
--- a/backend/src/database/migrations/2026_04_24_100000_create_ligas_table.php
+++ b/backend/src/database/migrations/2026_04_24_100000_create_ligas_table.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::create('ligas', function (Blueprint $table): void {
+            $table->id();
+            $table->foreignId('user_id')->constrained()->cascadeOnDelete();
+            $table->unsignedInteger('points')->default(0);
+            $table->timestamps();
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::dropIfExists('ligas');
+    }
+};


AlejandroJarque (approved):

laisrod (approved):
The migration looks clean. Just confirming — the missing fields (language, liga_type, week, year) are intentional for this sprint since we're starting with the base liga only, right? Will they be added in a future migration?

Thanks!

kenan-it-academy (approved):
This PR is okay, but it's weird that this is on it's own, it smells of not properly vertically splitting the tasks.

fdesouzabcn:
Hi @laisrod, 

Yes, at the sprint as discussed with Luis and @kenan-it-academy  the focus is on General Ranking only. Therefore it will return the minimal required fields, therefore week, year, liga_status won't be part of this sprint since general ranking does make any distinguishing about Liga type, week and doesn't display any avatar either.However your comment made me realize that we also removed language from the initial planning because at that time wasn't clear how languages would be used in the Liga context. But I'll bring it it back to discussion on Monday and apply a new PR toninckude it, if needed. I will get back to you on it this Monday morning.

fdesouzabcn:
@kenan-it-academy 

Your comment is right in the spot. We realized some challenges by not doing it properly in the beginning. We discussed it already on Friday when the work was already in progress, and decided to regroup on Monday morning to tackle it. Thanks for point out too.and if yo have any suggestion, ease let us know.

fdesouzabcn:
Note: I'm Not deleting the branch yet since we might need to add the language field on Monday.

fdesouzabcn:
Branch can be deleted as the language's exclusion from this sprint is confirmed.

PR #372, opened 2026-04-24
Feature/338 ticketing index returns assigned tickets

AlexBeltran31:
During the implementation of role-based access control for the ticket system, several inconsistencies were identified that could cause unexpected behavior:

✘ Assignee not included in ticket listing:
show() allows assignees to view a ticket they are assigned to, but index() only returns tickets where the user is the creator. An assignee can access the detail but the ticket never appears in their list.

The fix extends the filter to include tickets where the user is either the creator or the assignee (assignee_id), making the listing consistent with the existing access control logic. A test was added to cover this case and prevent regressions.


VictorCharneco (approved):
all good! the inconsistency between index() and show() is fixed and the test cover it well

AlejandroJarque (approved):

backend/src/tests/Feature/ReportingTickets/TicketControllerTest.php
```diff
@@ -649,6 +649,30 @@ public function any_authenticated_user_can_set_non_closed_status(): void{
         $this->assertDatabaseHas('tickets', ['id' => $ticket->id, 'status' => 'in_progress']);
     }
 
+    /** @test */
+    public function assignee_can_see_assigned_ticket_in_index(): void
+    {
+        $assignee = $this->authenticateUserWithRole('mentor');
```
kenan-it-academy, line 655: HOW does this file have 600+ lines??????

Backlog: split this insanity lol.

kenan-it-academy (approved):
approved but please add backlog to split the damn file lol

miguelm-montano (approved):
All good! Solid test and the logic is consistent

PR #381, opened 2026-04-27
feat: add LigaFactory, LigaSeeder and register in DatabaseSeeder

### LINKED ISSUE #304: Populate ligas table with realistic test data for local dev and automated tests.
Creates the Liga factory for generating test records and the Liga seeder, which populates one entry per existing user with a randomised points value. Additionally we need to ensure that all users exist in the database before the Liga entries are created.

AlejandroJarque:
This PR adds the factory and seeder infrastructure for the Liga feature.

LigaFactory was created to generate test Liga entries with a random points value between 0 and 100, delegating user creation to UserFactory. LigaSeeder was added to populate the ligas table with one entry per existing user, skipping any user that already has an entry to avoid duplicates. Finally, LigaSeeder was registered in DatabaseSeeder immediately after UserSeeder to guarantee that all users exist before the seeder runs.

Tasks 6 and 7 were grouped into this PR together with task 5 because they form a single coherent unit — the factory is only useful if there is a seeder to use it, and the seeder has no effect unless it is registered. Splitting them across separate PRs would have left the codebase in an incomplete state.

### DIFF
--- a/backend/src/app/Models/Liga.php
+++ b/backend/src/app/Models/Liga.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+class Liga extends Model
+{
+    use HasFactory;
+
+    protected $fillable = [
+        'user_id',
+        'points',
+    ];
+
+    public function user(): BelongsTo
+    {
+        return $this->belongsTo(User::class);
+    }
+}
--- a/backend/src/database/factories/LigaFactory.php
+++ b/backend/src/database/factories/LigaFactory.php
@@ -14,7 +14,7 @@ class LigaFactory extends Factory
 {
     public function definition(): array
     {
-        return[
+        return [
             'user_id' => User::factory(),
             'points' => fake()->numberBetween(0, 100),
         ];
--- a/backend/src/database/factories/LigaFactory.php
+++ b/backend/src/database/factories/LigaFactory.php
@@ -1,0 +1,22 @@
+<?php 
+
+declare(strict_types=1);
+
+namespace Database\Factories;
+
+use App\Models\User;
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+/** 
+ * @extends \Illuminate\Database\Eloquent\Factories\Factory<\App\Models\Liga> 
+ */
+class LigaFactory extends Factory
+{
+    public function definition(): array
+    {
+        return[
+            'user_id' => User::factory(),
+            'points' => fake()->numberBetween(0, 100),
+        ];
+    }
+}
--- a/backend/src/database/seeders/LigaSeeder.php
+++ b/backend/src/database/seeders/LigaSeeder.php
@@ -20,5 +20,7 @@ class LigaSeeder extends Seeder
                 ]);
             }
         });
+
+        Liga::factory()->count(5)->create();
     }
 }
--- a/backend/src/database/seeders/LigaSeeder.php
+++ b/backend/src/database/seeders/LigaSeeder.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Database\Seeders;
+
+use App\Models\Liga;
+use App\Models\User;
+use Illuminate\Database\Seeder;
+
+class LigaSeeder extends Seeder
+{
+    public function run(): void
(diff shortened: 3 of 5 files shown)


AlejandroJarque:
faltara tester el modelo con los factory y seeders, estoy a la espera de mergear la tarea de Hebe y añadire los test en esta misma pr

backend/src/database/seeders/LigaSeeder.php
```diff
@@ -0,0 +1,24 @@
...
+
+class LigaSeeder extends Seeder
+{
+    public function run(): void
+    {
+        User::all()->each(function(User $user) {
+            if(! Liga::where('user_id', $user->id)->exists()) {
+                Liga::create([
+                    'user_id' => $user->id,
+                    'points' => rand(0, 100),
```
fdesouzabcn: To maintain consistency with Liga Factory and leverage Faker's reproducible seeding, IMHO would be better to use `fake()->numberBetween(0, 100)` instead of `rand(0, 100)`.

backend/src/database/factories/LigaFactory.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+/** 
+ * @extends \Illuminate\Database\Eloquent\Factories\Factory<\App\Models\Liga> 
+ */
+class LigaFactory extends Factory
+{
+    public function definition(): array
+    {
+        return[
```
fdesouzabcn: missing space `return [`

fdesouzabcn (commented):
@AlejandroJarque  Thanks for setting up the LigaFactory and LigaSeeder — the core logic is solid, structure is clean and the idempotency guard is a nice touch. 

I tested it locally and found a few things worth checking before merging:

**As you know, App\Models\Liga model is missing**
So I borrowed the model temporarily from branch 302-model-creation to test. With it present, the seeder runs without errors — so the logic itself is fine. But as it stands the PR can't work in isolation, I recomend in future situations inform it so that other collegaues can know what to do and how proceed with the review. 

**Seeder gives a false green when no users exist**
When running against an empty database, the seeder reports success but creates zero records — because User::all() returns an empty collection and the loop body never executes. This means the missing model isn't caught either. Worth noting so whoever runs it knows to seed users first (or run the full DatabaseSeeder).

Other minor observations:

* LigaSeeder uses rand(0, 100) while LigaFactory uses fake()->numberBetween(0, 100) — better to keep them consistent and use fake() in both places.
* The exists() + create() pattern fires two queries per user — Liga::firstOrCreate(['user_id' => $user->id], ...]) does the same thing in one.
* `return[` on line 17 of the factory is missing a space → return [.
* due the false green and if times allow, consider add a simple test asserting that the factory creates a valid record and the seeder populates the table  - it would give confidence this works correctly.

Let me know if you have any questions!

fdesouzabcn:
@AlejandroJarque perdona que sólo ahora he visto tu comentario sobre lo de mergear y añadir tests. Así q tu ya lo genia medio avisado por los comentarios y no lo había visto.

triflip (approved):
Looks good to me. The structure is clear and the approach makes sense. No blocking comments from my side.

fdesouzabcn (approved):
Approving this to keep things moving. The core functionality works, and all tests pass—great job addressing the model and tests so quickly! 🙌 

However, the LigaSeeder seems to run too early in DatabaseSeeder, resulting in an extra user being created after LigaSeeder finishes. Running `migrate:fresh --seed` results in 7 users and 6 liga records. I suggest opening a ticket to address this observation, along with the other minor issues previously mentioned, at a later time. It's not urgent, just something to keep in mind.

backend/src/app/Models/Liga.php
```diff
@@ -0,0 +1,24 @@
...
+    use HasFactory;
+
+    protected $fillable = [
+        'user_id',
+        'points',
+    ];
+
+    public function user(): BelongsTo
+    {
+        return $this->belongsTo(User::class);
```
kenan-it-academy, line 22: Okay, talk to @HebeStark because this relationship makes more sense to me than each user having a league. Which is the correct implementation?

kenan-it-academy (dismissed):

kenan-it-academy (commented):
Accidentally marked Request Changes instead of Comment, lol

backend/src/app/Models/Liga.php
AlejandroJarque: We will end this relationship, and we will correct any conflicts, making sure to leave this version.

kenan-it-academy (approved):

HebeStark (approved):
vale

PR #396, opened 2026-04-28
feat: retrieve project owner in index and show endpoints

### LINKED ISSUE #323: Backend: pending to retrieve project owner and send it back
### Objective
Return the project owner data in index() and show() endpoints

##
✘ Add user relationship to with() in index() and show()
✘ Include owner (id, name) in the response
✘ Add tests to verify owner is returned

VictorCharneco:
### Objective
Return the project owner data in index() and show() endpoints

##
✘ Add user relationship to with() in index() and show()
✘ Include owner (id, name) in the response
✘ Document swagger and added tests to verify owner

### DIFF
--- a/backend/src/app/Http/Controllers/ListProjectsController.php
+++ b/backend/src/app/Http/Controllers/ListProjectsController.php
@@ -38,6 +38,12 @@ class ListProjectsController extends Controller
      *               @OA\Property(property="limit_date_inscription", type="string", format="date", nullable=true, example="2025-12-31"),
      *               @OA\Property(property="dev_front_number", type="integer", nullable=true, example=2),
      *               @OA\Property(property="dev_back_number", type="integer", nullable=true, example=2),
+     *               @OA\Property(
+     *                   property="owner",
+     *                   type="object",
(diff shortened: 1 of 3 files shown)


H3llynx (approved):
Cool! Retrieving the owner's name in addition to the id will also make it smoother on the front side at the time to display the project's owner information wherever requested.

Inbi-Nav (approved):
Good work! Everything looks correct and the owner is now included in both index and show

Just a few small suggestions:
- Maybe handle the case where the user is null to avoid errors.
- The mapping logic is repeated in index() and show(), could be reused.

VictorCharneco:
> Good work! Everything looks correct and the owner is now included in both index and show
> 
> Just a few small suggestions:
> 
> * Maybe handle the case where the user is null to avoid errors.
> * The mapping logic is repeated in index() and show(), could be reused.

great Inbi!!  i've applied both suggestions, many thanks🙏🏼🙌

kenan-it-academy (approved):
Clean and to the point. Love it!

PR #427, opened 2026-04-30
#397 feat(ticketing): add useCreateTicketing hook with types, endpoint stub and tests

### LINKED ISSUE #397: Create hook to create new tickets
Create the function in hook file to create a new ticket

vrsirvent:
Add `ticketingTypes.ts` with all ticket-related types and interfaces (`IntCreateTicket`, `IntTicket`, union types for status, priority, affected app and function)
Add `endPointTickets.ts` as a stub for the `createTicket` API call (pending merge of the real implementation)
Add `useCreateTicketing` hook managing async state: `ticketing`, `isLoading` and `error`
Add unit tests covering the 3 essential cases: happy path, loading lifecycle, and error recovery

**Notes:** `endPointTickets.ts` is intentionally a stub. The real API implementation is pending merge from a parallel branch. The hook and its tests will not require changes once that merge happens, as the endpoint is fully mocked in tests.

### DIFF
--- a/frontend/src/hooks/__tests__/useCreateTicketing.test.ts
+++ b/frontend/src/hooks/__tests__/useCreateTicketing.test.ts
@@ -60,7 +60,10 @@ describe("useCreateTicketing", () => {
   it("should set isLoading to true during the request", async () => {
     let resolvePromise!: (value: IntTicket) => void;
     mockCreateTicket.mockImplementation(
-      () => new Promise((res) => { resolvePromise = res; }),
+      () =>
+        new Promise((res) => {
+          resolvePromise = res;
+        }),
     );
 
     const { result } = renderHook(() => useCreateTicketing());
@@ -79,13 +82,17 @@ describe("useCreateTicketing", () => {
   });
 
   it("should set error on failure and reset it on next success", async () => {
-    mockCreateTicket.mockRejectedValueOnce(new Error("Error 500: Internal Server Error"));
+    mockCreateTicket.mockRejectedValueOnce(
+      new Error("Error 500: Internal Server Error"),
+    );
     const { result } = renderHook(() => useCreateTicketing());
 
     void result.current.submitTicketing(mockTicketData);
 
     await waitFor(() => {
-      expect(result.current.error?.message).toBe("Error 500: Internal Server Error");
+      expect(result.current.error?.message).toBe(
+        "Error 500: Internal Server Error",
+      );
     });
 
     expect(result.current.ticketing).toBeNull();
--- a/frontend/src/hooks/__tests__/useCreateTicketing.test.ts
+++ b/frontend/src/hooks/__tests__/useCreateTicketing.test.ts
@@ -41,7 +41,7 @@ describe("useCreateTicketing", () => {
     mockCreateTicket.mockResolvedValue(mockTicketResponse);
     const { result } = renderHook(() => useCreateTicketing());
 
-    expect(result.current.ticketing).toBeNull();
+    expect(result.current.ticketing).toEqual([]);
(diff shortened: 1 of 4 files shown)


VictorCharneco (changes requested):
the frontend check fails bcause of an ESLint error in src/api/endPointTickets.ts
the _data param is marked as unused by the linter, since its a stub maybe just add // eslint-disable-next-line @typescript-eslint/no-unused-vars above the function to silent it?

H3llynx (changes requested):
Good job! A few things to address before I can approve:

- **Prettier formatting:** Front tests are failing due to a formatting issue. Please run `npx prettier --write .` and commit again. FYI: this needs to be done before any commit including a front modification. Also, make sure to `cd frontend` before running it.

- **`[ticketing, setTicketing]` state:** I understand you're willing to save a refetch after creation. Good idea, but the implementation needs a small fix: `ticketing` should be typed as `IntTicket[]` instead of `IntTicket | null`, and rather than replacing the state with just the newly created ticket, you should append it to the existing list:
  ```ts
  setTicketing(prev => [...prev, newTicketing])
  ```

- **Broader design consideration:** Since this hook is both handling the creation logic and apparently owning the ticket list state, it's taking on two responsibilities. If the ticket list is consumed in more than one place, that state will probably have to be moved into a context.

H3llynx (changes requested):
fix ticketing + setTicketing y listo

VictorCharneco (approved):
OK, all fixed and solid tests!

H3llynx (approved):

frontend/src/hooks/__tests__/useCreateTicketing.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  created_at: "2026-04-29T10:00:00Z",
+  updated_at: "2026-04-29T10:00:00Z",
+};
+
+describe("useCreateTicketing", () => {
+  beforeEach(() => {
+    mockCreateTicket.mockClear();
+  });
+
+  it("should initialize with empty state and resolve successfully", async () => {
```
kenan-it-academy: this should be two tests. The first with the first block of "expects" and the second with the second block of "expects". In general, if you have an "and" in a test title, you likely should be doing two separate tests.

frontend/src/hooks/__tests__/useCreateTicketing.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    expect(result.current.isLoading).toBe(true);
+
+    await act(async () => {
+      resolvePromise(mockTicketResponse);
+    });
+
+    expect(result.current.isLoading).toBe(false);
+  });
+
+  it("should set error on failure and reset it on next success", async () => {
```
kenan-it-academy: this is two different test cases as well.

kenan-it-academy (changes requested):

frontend/src/hooks/__tests__/useCreateTicketing.test.ts
vrsirvent: "should set error on failure": verifies that a rejected API call sets the error and leaves ticketing empty.
"should reset error on next success": triggers a prior failure as setup, then verifies a subsequent success clears the error.

frontend/src/hooks/__tests__/useCreateTicketing.test.ts
vrsirvent: "should initialize with empty state": checks only the hook's initial values, sync, no mock needed.
"should resolve successfully": checks that submitTicketing calls the API correctly and updates the state.

kenan-it-academy (approved):

PR #446, opened 2026-05-05
Add tickets list in tickets page

### LINKED ISSUE #445: Add list component into ticketing page

albertvallsbe:
Added the TicketList component to the Ticketing page and connected it to the useTicketingGetAll hook to display tickets from the API.
Updated types to align frontend data with backend response and ensure type safety.
Included basic rendering test to verify the page loads and displays the ticket list correctly.
imatge

### DIFF
--- a/frontend/src/api/__tests__/endPointTickets.test.ts
+++ b/frontend/src/api/__tests__/endPointTickets.test.ts
@@ -1,5 +1,6 @@
-import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
-import { fetchAllTickets } from "../endPointTickets";
+import { describe, expect, it, vi } from "vitest";
+
+import { ticketsEndpoint } from "../endPointTickets";
 
 vi.mock("../../config", () => ({
   API_URL: "http://localhost:3000",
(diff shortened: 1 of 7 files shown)


Inbi-Nav (approved):
Nice work
Nothing blocking from my side

H3llynx (approved):
Good job overall, and it’s great to see frontend and backend aligned from the start.
I was a bit surprised, though, that the existing fetchAllTickets function in endPointTicket.ts (which calls the very same endpoint) wasn’t readapted (headers) and reused in the hook. Is there a specific reason for that? Since it’s not used anywhere at the moment, it might be worth removing it. That said, for consistency, I would probably move the raw fetch logic into endPointTicket.ts, but that's up to you :).

Also, this might just be a typo, but I was a bit confused by the test description "renders ticketing create form," since there is no form (yet?) in TicketingPage.
Approved 👌 Would be cool to see it with some backend data for tomorrow's demo.

albertvallsbe:
@H3llynx 
Good point, well spotted. You're right that the duplication wasn't correct and the existing fetchAllTickets should have been removed.

We followed this approach to split the work across three PRs within the team.

The idea here was to keep the endpoint layer limited to route definitions and let the hook handle the axios call (headers, auth token, loading state and error handling), even if this is not the usual pattern, with the last one focusing on the visual component.

That said, I agree the leftover function is confusing and should be removed to keep the API access pattern consistent.

PR #544, opened 2026-05-11
Feature/475 fix contributor avatars

### LINKED ISSUE #475: Fix contributor avatars disappearing after page refresh
As a project visitor, I want to see the avatar of each contributor in the project listing and detail pages, so that I can visually identify who is part of each project without the avatars disappearing after a page refresh.

Problem

The avatar field exists in the User model and is correctly stored in the database after GitHub login. However, the endpoints that return contributor data (index, show, and getContributors) do not include the avatar_url field in their responses. As a result, the frontend can only display avatars from data already held in memory — they disappear as soon as the page is refreshed or the user logs out and back in.

Solution

Add avatar_url to the contributor data returned by index(), show(), and getContributors() in ListProjectsController. The User model already has a getAvatarUrlAttribute() accessor that builds the correct URL, so no model changes are needed.

Additionally, the frontend is currently displaying a hardcoded placeholder image instead of the user's real GitHub avatar. Once the avatar_url is included in the contributor response, the frontend will need to replace the hardcoded image with the value returned by the API.

AlexBeltran31:
As a project visitor, I want to see the real GitHub avatar of each contributor in the project listing and detail pages, so that I can visually identify who is part of each project without the avatars disappearing after a page refresh.

Problem

The avatar field exists in the User model and is correctly stored in the database after GitHub login. However, the three endpoints that return contributor data (index, show, and getContributors) were not including the avatar_url field in their responses.

As a result, the frontend had no way to display real GitHub avatars after a page refresh — it was generating placeholder images from ui-avatars.com using only the contributor's name, ignoring the real avatar stored in the database.

Solution

Backend: Added avatar_url to the contributor data returned by index(), show(), and getContributors() in ListProjectsController. The User model already had a getAvatarUrlAttribute() accessor that builds the correct URL, so no model changes were needed.

Frontend: Updated ApiProjectContributor type to include avatar_url: string | null, and updated useProjectContributors hook to use the real GitHub avatar when available, falling back to a ui-avatars.com placeholder only when avatar_url is null.


H3llynx (approved):
Now when a user created a project they can see their own avatar instead of the placeholder, good job!

HebeStark (approved):
Para el alcance de esta rama: la funcionalidad está completa y correcta.

HebeStark (approved):
Para el alcance de esta rama: la funcionalidad está completa y correcta.

H3llynx (approved):

kenan-it-academy (approved):

Issue #477, opened 2026-05-07
[TICKETING][FE] Implement ticket creation logic


triflip:
Goal: Implement the useCreateTicket hook to handle ticket creation from the frontend.
Tasks:

Implement createTicket function in endPointTickets.ts (currently throws Not implemented)
Connect it with the existing useCreateTicketing.ts hook
Handle loading, error and success states

Endpoint: POST /api/tickets
Acceptance criteria:

Hook returns { submitTicketing, isLoading, error }
On success, ticket is added to the list
On error, error message is set

PR #561, opened 2026-05-12
test: add createTicket error handling test

triflip:
Adds a dedicated test for createTicket error handling.

This test was separated from PR #517 following review feedback requesting separation of happy path and error handling tests.

Depends on #477 (create ticket implementation).

### DIFF
--- a/frontend/src/api/__tests__/endPointTickets.test.ts
+++ b/frontend/src/api/__tests__/endPointTickets.test.ts
@@ -52,4 +52,9 @@ describe("createTicket", () => {
       }),
     );
   });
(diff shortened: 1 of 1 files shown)


vrsirvent (approved):
Correct, the error simulation and exception verification are adequate.

albertvallsbe:
If I am not mistaken, the test PR should always target `develop`, even if it is complementary or dependent on another PR.

Merging one feature branch into another can make the final PR too large, because it ends up including changes from both branches at the same time instead of keeping the reviews isolated and focused.

@kenan-it-academy Is this the approach recommended by the workflow theory for related PRs?

PR #562, opened 2026-05-12
Add status rules to createticketrequest

### LINKED ISSUE #560: [TICKETING] Add status and priority validation rules to CreateTicketRequest
### **Context**

Following the previous PR #527 that made non-essential fields `nullable`, `status` and `priority` were missing from `CreateTicketRequest` validation rules and from the defaults in `TicketController::store()`. Without these rules, the fields are not validated on input and status has no explicit default in the controller.

### **Criteria**

status and priority added as nullable in `CreateTicketRequest` using enum values
status defaults to `TicketStatusEnum::Pending` in `TicketController::store()`
priority remains null on creation as it is already nullable in the database
Tests updated to cover the new validation rules

### **To Do**

- Add status and priority rules to CreateTicketRequest
- Add status default in TicketController::store() with TicketStatusEnum import
- Update tests to reflect new validation rules if needed

miguelm-montano:
### **Summary**

 - Adds priority as a nullable validation rule in CreateTicketRequest using enum values, allows users to indicate urgency at creation time
 - status is not included in `CreateTicketRequest` it is always forced to TicketStatusEnum::Pending in TicketController@store() and cannot be set by the client
 - priority defaults to null on creation — the frontend displays null as "Baja" via the existing priority ?? "low" fallback, consistent with the API contract priority: TicketPriority | null and existing tickets prior to PR #562

### **Test plan**

- [x] php artisan test full suite passes inside Docker
- [x] TicketControllerTest - all tests pass
- [x] TicketRequestValidationTest - all tests pass with updated rules
- [x] TicketRequestsTest - all tests pass with updated rules
- [x] Creating a ticket with only description returns 201 with status set to pending and priority set to null

### DIFF
--- a/backend/src/app/Http/Controllers/Tickets/TicketController.php
+++ b/backend/src/app/Http/Controllers/Tickets/TicketController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers\Tickets;
 
 use App\Enums\AffectedAppEnum;
 use App\Enums\AffectedFunctionEnum;
+use App\Enums\TicketStatusEnum;
 use App\Enums\TicketTypeEnum;
 use App\Http\Controllers\Controller;
 use App\Models\Ticket;
@@ -64,6 +65,7 @@ class TicketController extends Controller
         $data['affected_app'] = $data['affected_app'] ?? AffectedAppEnum::Other->value;
         $data['type'] = $data['type'] ?? TicketTypeEnum::Error->value;
         $data['affected_function'] = $data['affected_function'] ?? AffectedFunctionEnum::Other->value;
+        $data['status'] = $data['status'] ?? TicketStatusEnum::Pending->value;
 
         $ticket = Ticket::create($data);
 
--- a/backend/src/app/Http/Controllers/Tickets/TicketController.php
+++ b/backend/src/app/Http/Controllers/Tickets/TicketController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers\Tickets;
 
 use App\Enums\AffectedAppEnum;
 use App\Enums\AffectedFunctionEnum;
+use App\Enums\TicketPriorityEnum;
 use App\Enums\TicketStatusEnum;
 use App\Enums\TicketTypeEnum;
 use App\Http\Controllers\Controller;
@@ -66,6 +67,7 @@ class TicketController extends Controller
         $data['type'] = $data['type'] ?? TicketTypeEnum::Error->value;
         $data['affected_function'] = $data['affected_function'] ?? AffectedFunctionEnum::Other->value;
         $data['status'] = $data['status'] ?? TicketStatusEnum::Pending->value;
+        $data['priority'] = $data['priority'] ?? TicketPriorityEnum::Low->value;
 
         $ticket = Ticket::create($data);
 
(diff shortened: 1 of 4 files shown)


triflip (approved):
Los cambios son coherentes con el issue #560 y con la lógica del ticketing. Las reglas nullable y los defaults en el controller funcionan correctamente, y los tests actualizados lo validan.

fdesouzabcn (commented):
@miguelm-montano 

Nice work on this PR! The implementation is clean and follows the existing patterns well.

A few minor observations I'd love to get your point of view on before I can approve.

**Status bypass on creation**
Since `status `is now a validated field in `CreateTicketRequest`, any user can currently POST `{"description": "x", "status": "closed"}` and skip the `updateStatus()` flow entirely — which means` closed_by` and `closed_at` would be null. Have you considered making the assignment unconditional in `store()`?
```
// instead of ??
$data['status'] = TicketStatusEnum::Pending->value;
```


**Priority default vs. ticket criteria**

The original ticket #560 states that priority should remain `null `on creation. Looking at the Figma mockup, it does show `"Baja"` (Low) as the displayed priority value, so my understanding is that the intention was always to default to Low — but the ticket itself wasn't updated to reflect that. If that's the case, it would be worth updating the ticket criteria before merging so the decision is documented and visible to everyone. Either way, let's ensure to align before this goes in.


**Priority on creation — role check**

Please correct if i'm undersatanding it wrong, but seems to me that `updatePriority()` would currently be admin/superadmin-only, but with this PR seems that any authenticated user would be capable to set priority at creation time (if functionality is made available in the FrontEnd). I beleive that would be worth a double-check if it's actually is that intentional? If not, you may want to strip priority from the data in store() unless the user has the right role.

Please let me know your thoughts and I can approve it.

fdesouzabcn:
@miguelm-montano, following up on my earlier comment about priority

I've been looking at this alongside Laís's PR #511 and I think there may be a cross-PR alignment issue worth sorting before we merge either. Laís's PR #511 builds on that assumption by adding a display fallback (ticket.priority ?? "low") that shows "Baixa" in the UI when the backend sends null — rather than storing "low" explicitly in the database.

If that's the intended contract, the line below would be changing the API behaviour that PR #511 was written against:
`$data['priority'] = $data['priority'] ?? TicketPriorityEnum::Low->value;`

Please correct me if I'm reading this wrong, but my understanding is that the backend should store null and the frontend handles the display — rather than the backend forcing a default. Either way, I'd love to get us all on the same page before either PR goes in. Would it make sense to loop Laís into this thread to confirm the intended contract?

laisrod (commented):
Hola Miguel, la PR está bien enfocada. Un punto:

Añadir status como campo nullable en CreateTicketRequest permite que cualquier usuario autenticado cree un ticket directamente con status: 'closed' o status: 'blocked', saltándose los endpoints dedicados (PATCH /tickets/{id}/status). Es esto intencionado, o debería eliminarse status de las reglas de creación y dejarlo solo como pending por defecto sin posibilidad de ser enviado por el cliente?

priority en cambio tiene más sentido — un usuario podría querer indicar la urgencia al crear el ticket.

miguelm-montano:
@laisrod @fdesouzabcn Thanks for the comments!! 

After the previous round of review feedback, the line `$data['priority'] = $data['priority'] ??  TicketPriorityEnum::Low->value` has already been removed from this PR. The current implementation uses `unset($data['priority'])` in `store()`, which means priority is stored as null in the database on creation, exactly the contract that PR #511 is written against.

So both PRs should now be aligned: backend stores null, frontend displays "Baixa" via the priority ?? "low" fallback. No behaviour change to the API contract.

That said, looping in Laís to confirm is a good idea, happy to do that if it helps close the loop before either PR is merged.

fdesouzabcn (approved):
Approved after team alignment and required changes applied.

laisrod (approved):

kenan-it-academy (approved):

PR #564, opened 2026-05-12
Enabled switcher between views

### LINKED ISSUE #485: Switch between league and general ranking

H3llynx:
### Purpose
Add and configure the existing switcher to the feature's dedicated page so users can toggle between league and global ranking content.

### Changes
- Extracted league-related code into a dedicated `WeeklyRanking` component.
- Added an (almost) empty `GlobalRanking` component, ready to be filled in when that content is available.
- Updated `LeaguesPage` so it now renders both components with the switcher to toggle between them.

### Testing:
- `WeeklyRanking.test.tsx` updated (no logic was changed, just moved from the page to the component since the page is not handling that logic anymore).
- New tests added in `LeaguesPage` to verify each view renders correctly on toggle.


albertvallsbe (approved):

Inbi-Nav (changes requested):
Need the following changes to avoid errors

- WeeklyRanking.test.tsx uses 'weekly_points' but 
in the backend returns 'points_weekly'. Verify the field name matches 
the actual API response. 

The rest LGTM

H3llynx:
@Inbi-Nav Thanks for flagging this, we actually had the reverted name in a few other files so I corrected them all now since we'll need the correct name for the rest to work. Requesting new review pls

Inbi-Nav (approved):

kenan-it-academy (approved):

PR #577, opened 2026-05-14
League filtering logic

H3llynx:
**What:** created the logic to allow display of results by league from the global ranking table provided by the backend.
**Files:**
- `utils/leagueUtils.ts` : filtering helper
- `useGlobakRanking.ts`: useMemo providing the new per league ranking
- `useGlobakRanking.text.ts`: tests confirming the logic actually works
 
**NB**:  a globalRanking placeholder state has been placed so far in `useGlobakRanking.ts`. The retrieval logic of the global ranking from the backend actually exists and is pending of approval in PR [#535](https://github.com/IT-Academy-BCN/ita-wiki-monorepo/pull/535)
Also, the PR seems a bit long but I actually wanted to have at least 3 entries in the mockData for proper testing.

PR #535, opened 2026-05-10
Retrieved global ranking from the backend

H3llynx:
**Changes:**
- Updated `config.ts` with the new league endpoints
- Created `endPointLeagues.ts `with the fetch logic
- Created `useGlobalRanking.ts` hook to retrieve the data and handle the loading / error logic

**Testing** in `useGlobalRanking.test.ts`

PR #578, opened 2026-05-14
League display 

### LINKED ISSUE #485: Switch between league and general ranking

H3llynx:
## What
Display each league's ranking grouped by league (Oro, Plata, Bronze), rendering a `LeagueList` per group with the correct label.

## Changes
- `WeeklyRanking.tsx` now maps over the grouped league data shape `[string, Omit<Ranking, 'points_weekly'>[]][]`, rendering a section per league with its label and standings
- since the hook and the endpoints are not yet merged, I updated `leagueService.ts` so it returns mocked data in the grouped format, mapping `points_weekly` → `points` and stripping `points_weekly` from entries. That mocked data is necessary to display something something right now and takes a lot of lines...
- Updated / fixed all affected tests to reflect the new data shape (tuple iteration instead of flat array).

## Preview:
Capture d’écran 2026-05-14 à 14 10 18

## Related PRs:
 - [#577 ](https://github.com/IT-Academy-BCN/ita-wiki-monorepo/pull/577) : where the filtering logic is defined to then apply to real backend data
 - [#535](https://github.com/IT-Academy-BCN/ita-wiki-monorepo/pull/535): where the data is actually retrieved from the backend


HebeStark (approved):
Cumple su objetivo principal: WeeklyRanking itera correctamente el formato agrupado y renderiza una sección por liga con su etiqueta.

albertvallsbe (approved):
Good job overall, although the PR may be slightly too large since it mixes data structure changes, UI updates and test adjustments.

The following test may not be necessary: 
This test is already covered in the LeagueList tests:
test("each item should have position, user_id and points", async () => {

This test also feels somewhat out of scope for the current feature:
test("each item should have created_at and updated_at", async () => {

There seems to be a small typo in the league labels. "Oro" should be "Or"

Why was points_weekly removed from the Liga type?

H3llynx:
Thanks @albertvallsbe. I removed those tests to lighten the PR a bit.

It's a big diff, I know, but the data isn't being shaped here, I just needed the mocked data adapted so the component is ready to merge. And without updating the tests, they fail since they were written against the old data structure.

Regarding removing points_weekly from Liga, I did it to adapt to what the list would actually show (so, "points" - weekly or not depending on the case) but you're right it's making a mess with how we are working right now so restablished it and readapted LeagueListProps instead.

HebeStark (approved):
creo que tambien deberías corregir también son el bug del fallback { id } → id y añadir la liga "3" al mock.

albertvallsbe (approved):

H3llynx:
no longer needed, data now retrieved from be

PR #579, opened 2026-05-14
Global ranking display

### LINKED ISSUE #485: Switch between league and general ranking

H3llynx:
## What
- Assembled `GlobalRanking.tsx` to show both the global ranking and the add points form.
- Added corresponding UI tests.

## Preview:
Capture d’écran 2026-05-14 à 16 01 00

## Related PRs:
 - [#535](https://github.com/IT-Academy-BCN/ita-wiki-monorepo/pull/535): where the data to insert in the ranking (list) component is retrieved from the backend

### DIFF
--- a/frontend/src/components/Leagues/__test__/GlobalRanking.test.tsx
+++ b/frontend/src/components/Leagues/__test__/GlobalRanking.test.tsx
@@ -1,7 +1,7 @@
 import "@testing-library/jest-dom/vitest";
 import { render, screen } from "@testing-library/react";
 import { describe, expect, it } from "vitest";
-import { GlobalRanking } from "../GlobalRanking";
+import { GlobalRanking } from "../../Leagues/GlobalRanking";
 
 describe("GlobalRanking", () => {
   it("renders the global ranking", () => {
--- a/frontend/src/components/Leagues/__test__/GlobalRanking.test.tsx
+++ b/frontend/src/components/Leagues/__test__/GlobalRanking.test.tsx
@@ -1,7 +1,7 @@
 import "@testing-library/jest-dom/vitest";
 import { render, screen } from "@testing-library/react";
 import { describe, expect, it } from "vitest";
-import { GlobalRanking } from "../../Leagues/GlobalRanking";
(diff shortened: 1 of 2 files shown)


fdesouzabcn (commented):
Hi @H3llynx ! 

The component composition looks clean and straightforward — easy to follow. I'm not a frontend expert so please take these as genuine questions rather than blockers. I just want to make sure I'm reading things correctly before approving.

#### Test file location
The new test landed at frontend/src/components/Leagues/__test__/GlobalRanking.test.tsx (capital L), but the component itself lives under frontend/src/components/leagues/ (lowercase l). I also noticed that the other tests for the same folder — AddLeaguePoints.test.tsx, LeagueList.test.tsx, WeeklyRanking.test.tsx — are all in frontend/src/components/leagues/__test__/. Would it make sense to move GlobalRanking.test.tsx there as well to stay consistent? Additionally, I'm wondering whether a case-sensitive filesystem might treat Leagues/ and leagues/ as two separate directories.

#### Hardcoded mock data
The component currently uses hardcoded entries (Jordi, Laia, Marc) rather than data from the backend. I know PR #535 already handles the data retrieval — is the plan to wire that up in a follow-up PR? If so, would it be possible to mention that in the PR description so the intent is clear to reviewers? The task description says "Make sure each wrapper displays the correct ranking/league information", so I just want to confirm this is on the radar.

Happy to approve once I you can review it and I can get more clarity on these points  — great work !

H3llynx:
hi @fdesouzabcn thank you for the review. Turns out I'll have to close this because this job was finally included in another PR but here's my replies to your question:

**Test file location**: we have a huge mess here because here in gh, there are 2 folders (Leagues and leagues) who are treated as separate. That causes quite a mess since our vscode handles everything as the same folder. Albert created a completely new one we'll put all the components into when it's approved, which will allow us to clean the repo).

**Hardcoded mock data**: correct, since we are still pending for the PR with the endpoint and the retrieval logic to be approved and merge, we have to deal with mocked data here in the meantime. I'm sorry I was not explicit enough about this in the description and thanks for letting me know so I'll correct that in next ones.

Thanks again!

PR #580, opened 2026-05-14
Aligned LeagueList component with what will be returned by the backend

### LINKED ISSUE #488: Create League list
Create the list element that will be used to display the general raking and each league's ranking.

Image

H3llynx:
**Changes:** 
- updated `LeagueList.tsx` component so it displays status and language, as per what the backend will returned
- readapted existing test

**Preview:** 
Capture d’écran 2026-05-15 à 17 38 44


vrsirvent (changes requested):
Suggestions::

LeagueList.tsx
The table layout is misconfigured: it has more columns than the CSS accounts for, which misaligns headers and rows (see grid-cols).

LeagueList.test.tsx 
The test that verifies column titles doesn't cover the two new ones added ("Estatus" and "Llenguatge").

H3llynx:
@vrsirvent thanks you're completely right I did not update the grid shape hence the issue. Now fixed. Regarding the test, I've added the new column headers but did not add any other test since I believe only happy path should be tested in this PR.

albertvallsbe (approved):

vrsirvent (approved):
Good job!

kenan-it-academy (approved):

PR #586, opened 2026-05-15
Connect de component to AddLeaguePoints with hook and endPoints

albertvallsbe:
Connected the AddLeaguePoints component with the add points hook and endpoint so users can submit point updates from the ranking UI.

Updated GlobalRanking to pass the loaded ranking users into AddLeaguePoints using the new user_id-based contract.

Adjusted the related tests to cover the new component integration and updated league navigation references.


H3llynx (approved):
Awesome to see the dropdown is now retrieving the users directly from the ranking. Hope we don’t get an out of scope for that, but happy to see the link to the feature is working again.
Only 1 recommendation here: add a description and maybe a more descriptive title to the PR :)

vrsirvent (approved):

kenan-it-academy (approved):
I feel like GlobalRanking and AddLeaguePoints could've been separated, but otherwise it's okay!

PR #588, opened 2026-05-18
Remove .env requirement and replace with ENV variable existence

kenan-it-academy:
For AWS we cannot always have a check for the .env file, as env variables will come directly.

Objectives:
- Remove any requirements for .env files existing
- Still allow current development workflow using .env files
- Ensure we support direct ENV variable injections from AWS

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -25,7 +25,6 @@ COPY --from=composer:latest /usr/bin/composer /usr/local/bin/composer
 COPY bootstrap ./bootstrap
 COPY public ./public
 COPY src ./src
-COPY .env.dev ./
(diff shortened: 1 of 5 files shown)


albertvallsbe (approved):

gregoridelrio (approved):
Tested locally — ran docker compose build and docker compose up -d from this branch. All 4 containers start correctly and the app loads fine at http://localhost/. Development workflow is not affected. The changes make sense — removing the hard dependency on .env files allows the app to work with direct environment variable injection from AWS. Approved!

orlandorojasoyarzun (approved):
Clean approach: removing the hard dependency on .env files while keeping local dev compatibility is what's needed for deployment with aws. Wiki tested locally and works as expected !

PR #629, opened 2026-05-20
Leagues - added correct fetch for league rankings

### LINKED ISSUE #607: Show the weekly ponctuation for each participants
- Create fetch with the ligas endpoint
- Create hook to retrieve the data

**Once the above is reviewed and merged:**
- Update the weekly ranking component so it displays that information
- Clean up: remove the useless grouping logic from the useGLobalRanking hook
- If time: fix the ranking color code

H3llynx:
### Added dedicated fetch for league ranking

**Changes:**
- Added missing fetch function to retrieve the league specific franking, since there is a dedicated endpoint for it
- Update the Liga type to match what's actually being returned
- Added corresponding test

**NB:** the PR is over > 50 lines because 52 lines are basically the single test for the fetch function.

### DIFF
--- a/frontend/src/api/__tests__/endPointLeagues.test.ts
+++ b/frontend/src/api/__tests__/endPointLeagues.test.ts
@@ -25,19 +25,18 @@ describe("fetchLeagueRanking", () => {
   });
 
   it("Should fetch the league rankings", async () => {
-    const mockResponse: LigaResponse = [
-      {
-        "1": [
-          {
-            position: 1,
-            user_id: 7,
-            username: "ckoelpin",
-            points_weekly: 99,
-            league_id: 1,
-          },
-        ],
-      },
-    ];
+    const mockResponse: LigaResponse = {
+      "1": [
+        {
+          position: 1,
+          user_id: 7,
+          username: "ckoelpin",
+          points_weekly: 99,
+          league_id: 1,
+        },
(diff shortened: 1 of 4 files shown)


Inbi-Nav (changes requested):
Almost there! Just a little change
- Test mock uses "http://localhost:8000/ligas" but 
   API_URL is "http://localhost/api/"

H3llynx:
Thanks for flagging @Inbi-Nav. I've just updated it in my test. However @kenan-it-academy FYI there are several tests using `http://localhost:8000` as a api url mock. If mocks now have to align with backend data maybe we should update them all?

claudiabcn (approved):
PR correcta. Pocas líneas, y cumple con lo que se pide.

Inbi-Nav:
> Thanks for flagging @Inbi-Nav. I've just updated it in my test. However @kenan-it-academy FYI there are several tests using `http://localhost:8000` as a api url mock. If mocks now have to align with backend data maybe we should update them all?

Lets' update them later. For now, it's fine as it is

Inbi-Nav (approved):

kenan-it-academy:
Everything is mocked in tests anyway, it could be `http://chicken.caw/` for all I care, lol

kenan-it-academy (approved):

PR #647, opened 2026-05-21
Add creator role to ticket index response for admin users

### LINKED ISSUE #605: [TICKETING] As a developer, I want to see the role of the ticket creator

DESCRIPTION

The system currently does not display the role of the user who creates a ticket.
This task requires adding that information to the ticket detail view.

OUT OF SCOPE

- Layout, CSS

ACCEPTANCE CRITERIA

- The ticket response includes the creator's role only when the authenticated user has the admin role.
- A non-admin user does not see the creator's role.
- Test for both cases
- Role field will be added to the existing ticket detail view

gregoridelrio:
When an authenticated user with the admin role retrieves the ticket list, the response now includes the role field of the ticket creator inside the code_connect object. Non-admin users will not see this field.

The admin role is being used as a temporary replacement for the developer role, which doesn't exist yet. Once the developer role is created, it will replace admin here. The superadmin is intentionally excluded from this feature for now. This was agreed with the team.

### DIFF
--- a/backend/src/app/Http/Controllers/Tickets/TicketController.php
+++ b/backend/src/app/Http/Controllers/Tickets/TicketController.php
@@ -30,7 +30,7 @@ class TicketController extends Controller
 
         $tickets = $query->get();
 
-        if (auth()->user()->hasAnyRole(['admin'])) {
+        if (auth()->user()->hasAnyRole(['admin' , 'superadmin'])) {
             $tickets->each(function ($ticket) {
                 if ($ticket->codeConnect) {
                     $ticket->codeConnect->role = $ticket->codeConnect->getRoleName();
--- a/backend/src/app/Http/Controllers/Tickets/TicketController.php
+++ b/backend/src/app/Http/Controllers/Tickets/TicketController.php
@@ -28,9 +28,20 @@ class TicketController extends Controller
             });
         }
 
+        $tickets = $query->get();
+
+        if (auth()->user()->hasAnyRole(['admin'])) {
+            $tickets->each(function ($ticket) {
+                if ($ticket->codeConnect) {
+                    $ticket->codeConnect->role = $ticket->codeConnect->getRoleName();
+                    $ticket->codeConnect->makeHidden('roles');
+                }
+            });
+        }
+
         return response()->json([
             'success' => true,
-            'data' => $query->get()
+            'data' => $tickets
         ]);
     }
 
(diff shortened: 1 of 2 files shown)


miguelm-montano (commented):
The implementation is good!

Both tests cover exactly the described behavior:

- `admin_can_see_creator_role_in_ticket_list` verifies that the role key exists
- `non_admin_cannot_see_creator_role_in_ticket_list` verifies that this key does not exist for a user without the admin role

Before approving:

The condition uses `hasAnyRole(['admin'])` but the description mentions "**admin users**". The controller consistently uses `['admin', 'superadmin']`. Is there any reason why `['superadmin']` is not included here?

Let me know, and I'll approve it!

gregoridelrio:
Good point! However, this is intentional. The admin role is being used as a temporary replacement for the developer role, which doesn't exist yet. Once the developer role is created, it will replace admin here. The superadmin is intentionally excluded from this feature for now. This was agreed with the team.

vrsirvent (changes requested):
In the ticketController.php file, you've only added “admin” to the condition; in the rest of the code, “superadmin” is also added.

Check the codeconnect role in the line $query = Ticket::with ...

miguelm-montano:
@gregoridelrio Great, thanks for clarifying that.
Could you include this information in the PR description? 

Also, If you're going to replace “admin” with ‘developer’ later, what happens to “superadmin” then?

claudiabcn (approved):
Short PR. It is good, the PR does what is expected should do.

gregoridelrio:
Updated! Added superadmin to the condition so both admin and superadmin can see the creator's role, consistent with the rest of the controller. Also added a test covering the superadmin case. Also updated the PR description to clarify that admin is a temporary replacement for the developer role, which will be created in a future task.

orlandorojasoyarzun (approved):
Great job ! The feedback was handled correctly !
 
- Adding superadmin to the condition to match the rest of the controller
- Adding a new test to cover the superadmin case
- Updating the PR description to clarify the temporary nature of the admin role

It's clear you paid attention to what both @miguelm-montano and @vrsirvent pointed out, and you addressed everything in a clean and organiced way. Very well !

Right in the limit of the scope with 52 lines edited though.

vrsirvent (approved):
That looks great! Nice work!

miguelm-montano:
Thanks for the changes @gregoridelrio 

Now `hasAnyRole` follows the pattern of the rest of the file and the tests cover the necessary cases 🙌🏻

kenan-it-academy (approved):

PR #651, opened 2026-05-21
feat(#645): add start_date field to project creation form

### LINKED ISSUE #645: Add start and end date fields to the project creation form
[ WHAT ]
Add start_date and end_date date fields to the project creation form (FormCreate.tsx).

[ TASKS ]

Add start_date input to the form (label in Catalan)
Register start_date in the form handler
Include start_date in the API request payload
Add end_date autocalculation from start_date + duration
Add minimal validation (end_date cannot be earlier than start_date)
Display validation error message in Catalan
[ PRs ]
This issue will be resolved with 3 small PRs :

PR1: Add start_date field + payload
PR2: end_date autocalculation
PR3: Date validation

triflip:
Adds start_date input field to the project creation form (FormCreate.tsx).

- Add start_date field to form state
- Register start_date in handleDeadLine handler
- Include start_date in API request payload
- Add start_date to form validation
- Test: renders start_date input and includes it in payload

Closes #645

### DIFF
--- a/frontend/src/components/code-connect/FormCreate.tsx
+++ b/frontend/src/components/code-connect/FormCreate.tsx
@@ -149,7 +149,7 @@ const FormCreate = () => {
       dev_back_number: formData.dev_back_number,
       time_duration: getTimeDuration(formData.time, formData.unitTime),
       limit_date_inscription: formData.limit_date_inscription,
-      ...(formData.start_date && { start_date: formData.start_date }),
+      start_date: formData.start_date,
     };
 
     try {
--- a/frontend/src/components/code-connect/FormCreate.tsx
+++ b/frontend/src/components/code-connect/FormCreate.tsx
@@ -31,6 +31,7 @@ const FormCreate = () => {
     time: 0,
     unitTime: "",
     limit_date_inscription: "",
+    start_date: "",
   });
   const [isSubmitting, setIsSubmitting] = useState(false);
   const [roadmap, setRoadmap] = useState<Task[]>([]);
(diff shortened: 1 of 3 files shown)


miguelm-montano (changes requested):
Everything matches the description you provided.

- The tests verify the field and the payload
- input with the field on catalan
- `start_date` in the state

The only thing that needs to be implemented is: 
In the **formPayload** of **handleSubmit**, `start_date` is enclosed in a conditional spread: 
```
(formData.start_date && { start_date: formData.start_date })
```
                                                                                
This means that if `start_date` is empty, it won't be included in the payload, but the backend requires it, so it would return a 422 error. Since it's already validated in
`validateForm`, you can simplify it to:                                                                                                            
```
start_date: formData.start_date
```

That would be the only thing to change.

triflip:
Done! Simplified the start_date payload as suggested — removed the conditional spread since validateForm already ensures start_date is not empty.

miguelm-montano (approved):
Thanks for the changes.
Everything looks good! Nice work! 🙌🏻

vrsirvent (changes requested):
start_date is defined as optional in types.ts (start_date?: string) but the form treats it as required: it is validated before submission and always included in the payload. Remove the ? so the type reflects the actual behavior:
types.ts
start_date: string;

triflip:
Done! Removed the ? from start_date in IntCodeConnect — the type now reflects the actual behavior of the form.

vrsirvent (approved):
That looks great! Nice work!

kenan-it-academy (approved):

PR #653, opened 2026-05-21
Leagues - implemented a button for each scoring activity

### LINKED ISSUE #648: Create predefined activity buttons
Create predefined activity buttons for mentors to add points 
to students quickly without entering a manual amount.

H3llynx:
### Created a button for each scoring activity

**Changes:**
As part of this sprint goal, the AddLeaguePoints component has been refactored to include a dedicated button for each scoring activity.

This includes:
- Mapping scoring activities and their associated points via `pointSystem`
- Rendering a button for each activity and ensuring the correct points value is passed on submission
- Partial refactor of the useAddLeaguePoints.ts hook to support the updated structure.

**Preview:**
Capture d’écran 2026-05-21 à 15 06 17

**Notes**:
- The backend currently applies a default value of 5 points per submission. As a temporary measure, all actions therefore result in 5 points being added on the backend, even though the front is ready to pass the correct points.
- Since CTAs are now less explicit, I added a title to the widget. That's the only UI improvement I added to avoid out of scope.

### DIFF
--- a/frontend/src/components/leagues-ranking/AddLeaguePoints/AddLeaguePoints.tsx
+++ b/frontend/src/components/leagues-ranking/AddLeaguePoints/AddLeaguePoints.tsx
@@ -16,8 +16,9 @@ export const AddLeaguePoints = ({
   users,
 }: AddLeaguePointsProps): JSX.Element => {
   const [selectedUsername, setSelectedUsername] = useState<string>("");
+  const [addedPoints, setAddedPoints] = useState<number | null>(null);
 
-  const { addPoints, error, isLoading } = useAddLeaguePoints({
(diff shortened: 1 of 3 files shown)


triflip (approved):
Clean and focused implementation.
The refactor makes sense for supporting dynamic scoring, and the UI change stays within scope.
Even if the PR is a bit large, it’s fully justified since all changes belong to the same feature.
Looks good to me!

orlandorojasoyarzun (approved):
I agree with the fact that the PR is a bit long and gets out of scope within amount of edited lines. 

The refactor is valid and it's well executed. The PR note about the backend limitation shows good awareness of the bigger picture. Nice one !

kenan-it-academy (approved):

PR #655, opened 2026-05-21
feat(liga): add weekly reset command for points_weekly

### LINKED ISSUE #549: Update function that resets points_weekly
Reset points_weekly to 0 for all users after promotions/demotions are calculated

orlandorojasoyarzun:
It has been implemented a Laravel command called Scheduler which automatically resets the weekly points for all users in the league system.

- Laravel scheduler will run a command every Sunday at midnight (configured in the backend folder routes/console.php).
- It's easy and accesible to modified if the Mentor would like to start the week either Monday or any other day.
- The command resets the points_weekly field to 0 for every user participating in the leagues.
- This makes sure each week starts fresh — students begin with zero weekly points and accumulate them as they collaborate. When Sunday arrives, the reset is applied, closing that week's cycle and opening a new one.

Testing: 
All 50 tests are passing, including the new one that verifies the reset command works correctly.

### DIFF
--- a/backend/src/routes/console.php
+++ b/backend/src/routes/console.php
@@ -2,7 +2,15 @@
 
 use Illuminate\Foundation\Inspiring;
 use Illuminate\Support\Facades\Artisan;
+use Illuminate\Support\Facades\Schedule;
 
 Artisan::command('inspire', function () {
     $this->comment(Inspiring::quote());
 })->purpose('Display an inspiring quote');
+
+Artisan::command('liga:reset-weekly', function () {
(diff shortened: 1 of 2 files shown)


H3llynx (approved):

gregoridelrio (approved):
Tested locally — ran the LigaResetWeeklyCommandTest and it passes. The command correctly resets points_weekly to 0 while keeping points intact. The scheduler is set to run every Sunday at midnight which makes sense for a weekly reset. The test covers both cases (weekly points reset, total points preserved). Clean implementation! Approved!

kenan-it-academy (approved):

PR #657, opened 2026-05-21
feature(#594): Add getComments API function

### LINKED ISSUE #594: [TICKETING] As a developer, I want to add comments to a specific ticket

DESCRIPTION

As an administrator, I want to add comments to a specific ticket in order to communicate updates, clarifications or resolutions directly within the context of the ticket.

OUT OF SCOPE

- Editing or deleting own comments
- Layout, CSS

ACCEPTANCE CRITERIA

- Existing comments are displayed on the ticket detail view
- Each comment shows the name of the user who wrote it and the date it was published
- There is a text field and a "Send" button to write and submit a new comment
- When a comment is submitted, an API call is made to save it
- The comment list updates after a successful API response
- An error message is displayed if the API call fails
-

vrsirvent:
Prepare the internal work so the application can fetch comments from the server for a specific ticket.

Changes:
- Defines the shape of a comment (text, author, date, etc.) so the application knows what to expect from the server (ticketingTypes.ts)
- Creates the server connection to fetch the comments of a ticket (endPointTickets.ts)
- Verifies that the connection works correctly (endPointTickets.test.ts)

### DIFF
--- a/frontend/src/api/__tests__/endPointTickets.test.ts
+++ b/frontend/src/api/__tests__/endPointTickets.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import axios from "axios";
-import { createTicket } from "../endPointTickets";
+import { createTicket, getComments } from "../endPointTickets";
 import { IntCreateTicket } from "../../types/ticketingTypes";
 
 vi.mock("axios");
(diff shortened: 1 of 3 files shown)


H3llynx (approved):
Good job

Inbi-Nav (changes requested):
The manual fag is not needed in this sprint:
Changes needed:
- Remove the manual flag from the request
- Remove the if (!$isManual) condition
- Always increment both points and points_weekly
- Also add validation to prevent negative or zero points

Inbi-Nav (approved):
LGTM
The endpoint, types and tests are aligned with the backend structure and the implementation is clear.

kenan-it-academy (approved):

Inbi-Nav (approved):

H3llynx (approved):

H3llynx (approved):

PR #659, opened 2026-05-22
feat: update LigaController to support manual points

### LINKED ISSUE #610: Update LigaController to support for manual points

HebeStark:
###Description:
Remove manual flag from addPoints endpoint and add input validation.

The manual flag is not needed in this sprint. Points always increment
both fields (points and points_weekly) as required by the weekly league
classification logic.

Added validation to reject zero or negative point values (min:1).

### DIFF
--- a/backend/src/app/Http/Controllers/LigaController.php
+++ b/backend/src/app/Http/Controllers/LigaController.php
@@ -74,17 +74,22 @@ class LigaController extends Controller
         }
 
         $pointsToAdd = $request->input('points', 5);
+        $isManual    = $request->boolean('manual', false);
 
         $entry->increment('points', $pointsToAdd);
-        $entry->increment('points_weekly', $pointsToAdd);
+
+        if (!$isManual) {
+            $entry->increment('points_weekly', $pointsToAdd);
+        }
 
         $entry->refresh();
 
         return response()->json([
-            'user_id' => $user->id,
-            'points'  => $entry->points,
+            'user_id'       => $user->id,
+            'points'        => $entry->points,
             'points_weekly' => $entry->points_weekly,
         ]);
     }
+
 }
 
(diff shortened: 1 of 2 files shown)


Inbi-Nav (changes requested):
The manual fag is not needed in this sprint:
Changes needed:
- Remove the manual flag from the request
- Remove the if (!$isManual) condition
- Always increment both points and points_weekly
- Also add validation to prevent negative or zero points

Inbi-Nav (approved):
LGTM! 

One suggestion: add a test for custom points amount 
to verify both points and points_weekly increment correctly 
when a specific value is passed.

H3llynx (commented):
Great! I guess I can now follow up with the frontend on that

HebeStark:
@Inbi-Nav  agregado el test

H3llynx (approved):

backend/src/app/Http/Controllers/LigaController.php
```diff
@@ -73,18 +73,23 @@ public function addPoints(Request $request, User $user): JsonResponse
...
         }
 
-        $pointsToAdd = $request->input('points', 5);
+        $validated = $request->validate([
+            'points' => 'nullable|integer|min:1',
+        ]);
 
-        $entry->increment('points', $pointsToAdd);
-        $entry->increment('points_weekly', $pointsToAdd);
+         $pointsToAdd = $validated['points'] ?? 5;
```
kenan-it-academy, line 85: I understand this default is for backwards compatibility, so make sure to add a backlog task to remove it once frontend is adapted!

kenan-it-academy (approved):

backend/src/app/Http/Controllers/LigaController.php
HebeStark: @kenan-it-academy Helene is already working on that so no need to add it to backlog

PR #666, opened 2026-05-22
League color fix

### LINKED ISSUE #607: Show the weekly ponctuation for each participants

H3llynx:
**Context**
Previously, the top 3 and last 3 participants were marked in green and red in both the global ranking and league tables, separately for each table.
During the last sprint planning, it was clarified that colors should indicate participants who can ascend or descend within leagues, not simply the top/bottom positions of each table.

**Changes**
- Updated the list component to accept optional showUp and showDown props.
- Disabled color highlighting in the global ranking.
- Enabled color highlighting in the weekly ranking (leagues) only when relevant: for example, the top 3 of the gold league cannot ascend, and the last 3 of the last league cannot descend, so no colors are shown in those cases.

### DIFF
--- a/frontend/src/components/leagues-ranking/LeagueList/LeagueList.tsx
+++ b/frontend/src/components/leagues-ranking/LeagueList/LeagueList.tsx
@@ -1,14 +1,18 @@
 import type { LeagueListProps } from "../../../types/league";
 
-export const LeagueList = ({ standings }: LeagueListProps) => {
+export const LeagueList = ({
+  standings,
+  showUp,
+  showDown,
+}: LeagueListProps) => {
   const getStyle = (index: number) => {
     const base =
       "grid grid-cols-6 border-b last:border-b-0 border-gray-400 px-4 py-4 text-sm text-center text-slate-950";
 
     const color =
-      standings.length > 3 && index < 3
+      showUp && standings.length > 3 && index < 3
         ? "bg-green-100"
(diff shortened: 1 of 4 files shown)


claudiabcn (approved):
That PR is okay. Approved.

orlandorojasoyarzun (approved):
Well done !

Tested the frontend from your branch and everything is working correctly from a code perspective. 

For visual testing: I noticed the seed data has very few participants per league (1-3), which means the conditions for showing colours (>3 for green, >6 for red) are never met. 

To see the colours in action, the LigaSeeder would need more participants. 

This is not a blocker, the implementation is solid !

Just something to keep in mind for future testing (or for the demo).

H3llynx:
@malmental 100% agreed. I actually had to test the visual with mocked data precisely for that reason. How can I increase the amount of participants returned by BE?

orlandorojasoyarzun:
@H3llynx  It has to be done in the data seeders coming from the backend, is something we could speak about next daily to prevent this and be able to show it visually ,)

kenan-it-academy (approved):

PR #669, opened 2026-05-23
feat: add league promotion and demotion processing

Inbi-Nav:
**Description**

Implemented the weekly league promotion and demotion system 

**Changes included**

- Added `ProcessLeaguePromotions` artisan command
- Implemented promotion/demotion logic between leagues
- Added protection for minimum and maximum league boundaries
- Prevented users from being promoted/demoted multiple times in the same execution
- Added `LeagueTypeEnum` to define league 
- Updated `Liga` model to use league enum casting
- Updated factories and seeders to use valid league values only
- Added feature tests covering

### DIFF
--- a/backend/bootstrap/app.php
+++ b/backend/bootstrap/app.php
@@ -12,6 +12,9 @@ return Application::configure(basePath: dirname(__DIR__))
         commands: __DIR__.'/../src/routes/console.php',
         health: '/up',
     )
+    ->withCommands([
+        \App\Console\Commands\ProcessLeaguePromotions::class,
+    ])
     ->withMiddleware(function (Middleware $middleware) {
         $middleware->api(prepend: [
             \Illuminate\Http\Middleware\HandleCors::class,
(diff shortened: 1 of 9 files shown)


orlandorojasoyarzun (commented):
Great job implementing the league promotion and demotion system ! The logic looks solid. The PR is very well described and defined. The enum approach for league types is also much cleaner than using raw integers.

I'd like to add: 
the PR is exceeding the Academy scope. Since this is a full feature tho and considering just the artisan command  is 50 lines already, consider splitting it into smaller pieces (for example: one PR for the enum, another for the command, another for tests). 

Great catch on the test fix!

This should be approved, but since last sessions with the team we spoke about splitting PR's I'd like to hear about @lxcxsito

lxcxsito (approved):
Great work on this PR. The implementation is clean and well-structured.

I also like the safeguards for league boundaries and the prevention of multiple promotions/demotions during the same execution. The feature tests cover the critical scenarios well and give good confidence in the behavior.

Everything looks good to me — approved.

orlandorojasoyarzun (approved):

kenan-it-academy (changes requested):
WOAH this is a huge PR! We'll talk tomorrow about splitting it lol

PR #672, opened 2026-05-26
Show ticket creator role in ticket list

### LINKED ISSUE #605: [TICKETING] As a developer, I want to see the role of the ticket creator

gregoridelrio:
Adds a new 'Rol' column to the ticket list that displays the role of the ticket creator. The data comes from the existing `code_connect.role` field returned by the backend for admin and superadmin users. Non-admin users will see "-".

A test has been added to verify the column header renders correctly.

Admin and SuperAdmin will see:

image

And other roles will see:

image


vrsirvent (changes requested):
The test does not verify that the role value is displayed correctly in the table. code_connect with a role should be added to the mock to verify that the value appears in the table.

gregoridelrio:
Added code_connect with role to the mock and a new test that verifies the role value is rendered in the table.

vrsirvent (approved):
That looks great! Nice work!

claudiabcn (approved):
✅ Approved.

Clean column addition — grid updated consistently in header and rows, nullable role handled with `?? "-"`, and tests cover the header and value rendering. No issues.

DanielVeraMunoz (approved):

kenan-it-academy (approved):

PR #673, opened 2026-05-26
Feature/593 ticketing: add DropdownMenu

### LINKED ISSUE #593: [TICKETING] As a developer, I want to change the status of a ticket

DESCRIPTION

OUT OF SCOPE
- Layout, CSS

ACCEPTANCE CRITERIA

claudiabcn:
This PR adds a generic DropdownMenu atom component that wraps DropdownButtonComponent with open/close state and click outside detection. DropdownButtonComponent has been extended with an optional className prop to allow margin overrides when used inside table cells.

### DIFF
--- a/frontend/src/components/atoms/DropdownButtonComponent.tsx
+++ b/frontend/src/components/atoms/DropdownButtonComponent.tsx
@@ -5,6 +5,7 @@ interface DropdownButtonComponentProps {
   onClick?: () => void;
   icon?: string;
   disabled?: boolean;
+  className?: string;
 }
 
 const DropdownButtonComponent: FC<DropdownButtonComponentProps> = ({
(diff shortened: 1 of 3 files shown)


H3llynx (approved):
Great job!
I honestly fail to see how this PR could shorter taking into account that the new component is minimalistic and the test just covers the expected behavior.

triflip (approved):
Nice job! The PR is a bit big, but totally justified since it adds a complete new atom plus all the tests.
The code is clear, the behaviour is solid, and everything looks consistent with the design system.
LGTM from my side.

kenan-it-academy (changes requested):
I'm sure you know why I'm not accepting this 😉

claudiabcn:
> I'm sure you know why I'm not accepting this 😉

Lo suyo seria que lo dijeras no? Adivina no soy. Por tu mensaje imagino que es larga. La dividido haciendo el test en una pr y el componente en otra pr?  O puedo no hacer el test? Agradecería un comentario más descriptivo de tu rechazo gracias. Si no es imposible saber que es lo que quieres, aun no tengo ese poder.

PR #680, opened 2026-05-27
Feature/593 ticketing: Editing DropdownButtonComponent

### LINKED ISSUE #593: [TICKETING] As a developer, I want to change the status of a ticket

claudiabcn:
DropdownButtonComponent has been extended with an optional className prop to allow margin overrides when used inside table cells. 
This change, the creation of the component, is also included in PR #690, but since this PR depends on it, it has been added here to avoid blocking the implementation.


lxcxsito (approved):
Very good , approved.

H3llynx (approved):
super short! approved

H3llynx (approved):
ok so I'l check DropdownMenu and its test in PR 690, as I see the only change here is between those 2 is the readaption of DropdownButtonComponent.
short and straightforward, approved.

miguelm-montano (approved):
- The main change is properly implemented in `DropdownButtonComponent.tsx`
- The duplicate `DropdownMenu.tsx` file is justified by the description; I think the conflict will be minimal when merging

Approved ✅

kenan-it-academy (approved):

PR #690, opened 2026-05-27
Feature/593 ticketing: add DropdownMenu as atom

### LINKED ISSUE #593: [TICKETING] As a developer, I want to change the status of a ticket

claudiabcn:
This PR adds a generic DropdownMenu atom component that wraps DropdownButtonComponent with open/close state and click outside detection. It includes unit tests covering render, open/close behavior, option selection and disabled state.
DropdownOption type has been moved from the component file to ticketingTypes.ts to make it reusable across the app.


H3llynx (approved):

miguelm-montano (approved):
From what I understand, everything is implemented correctly in the PR.

You created the component and moved the `DropdownOption` type to `ticketingTypes.ts`, which technically means it does two things, but in this case, from my perspective, it’s justified because it’s necessary for the component to be reusable. 

The tests work and the number of lines is due to the tests that evaluate the five necessary cases: render, open, select, close after selection and disabled state.

Beyond the size, which is justified by the tests, I think everything is implemented correctly.

kenan-it-academy (changes requested):
Way too many usecases in here!

1. Basic opening and selecting
2. Closing after selection
3. Disabled

lxcxsito (approved):
Tests have been simplified in this pull request. Approved

H3llynx (approved):

lxcxsito (approved):
Tests have been simplified in this pull request. Approved

claudiabcn:
> Way too many usecases in here!
> 
> 1. Basic opening and selecting
> 2. Closing after selection
> 3. Disabled

OK!! Right, Done. Please, reconsider :).

kenan-it-academy (commented):
Now I'm more confused: these are test for already-existing funcionality?

claudiabcn:
> Now I'm more confused: these are test for already-existing funcionality?

No... There was a PR related...https://github.com/IT-Academy-BCN/ita-wiki-monorepo/pull/680.

You approved now the https://github.com/IT-Academy-BCN/ita-wiki-monorepo/pull/680, then, the component started to exist. But the component and the test were here. But as the merge has done, then, now this PR is smaller, because of that. But it is the same.

kenan-it-academy (approved):
My bad for allowing the other PR in without tests. Won't happen again 😉

PR #691, opened 2026-05-27
Code Connect - Allow rejecting a pending join request

### LINKED ISSUE #619: Allow rejecting a pending join request
### **Context**

`PendingRequest` has an "**Acceptar**" button that calls `updateContributorStatus `with status accepted, but there is no way to **reject** a join request from the UI. The backend endpoint PATCH `/api/codeconnect/{id}/contributors/{id}/status` already supports rejected as a valid status.

### **Criteria**

- Each pending request displays a "**Rebutjar**" button alongside the existing "**Acceptar**" button
- Clicking the button calls `updateContributorStatus` with status rejected

### **To Do**

- Add "**Rebutjar**" button to `PendingRequest`
- Call `updateContributorStatus` with rejected on click

H3llynx:
### Created a separate “Reject” button component

**Context:**
The backend logic and endpoint for rejecting a contribution request already existed, but the UI button to trigger it was missing.

**Changes:**
Created a dedicated RejectButton component (and its test) with its own handleClick function that calls the reject endpoint.

**Notes:**
- This button is intended for use in the `PendingRequest` component. However, since `PendingRequest` hasn’t been merged yet, I implemented the button as a standalone component to avoid future merge conflicts. That means a follow-up PR is needed to integrate it into `PendingRequest`.
- The reject endpoint requires both the contributor ID and the project ID. To support this, I added a new `project` prop to PendingRequest.


miguelm-montano (changes requested):
Hélène, everything looks good!! The tests are well implemented and the PR does exactly what it needs to do.                                                                                  

The only minor thing I'd suggest is renaming the project prop in `PendingRequest.tsx` to `projectId` instead of `project` to stay consistent with the naming convention used across the rest of the project

After that, I'll approve it right away!

H3llynx:
@miguelm-montano makes total sense. Done

miguelm-montano (approved):
I can see the button clearly, nothing else to add, just it's great!! 🙌🏻

claudiabcn (changes requested):
Great work overall! The component structure is clean and the logic is straightforward. 
Required changes:
console.log(res) in handleReject — This should be removed before merging. 

Minor suggestions:
Hardcoded string in the button — "Rebutjar" is hardcoded.

H3llynx:
@claudiabcn ooops, forgot about that one thanks. Done.
I agree that hardcoded text is not the best option (mostly if the app is supposed to be available in several languages at some point), but I'd leave this for another sprint (are we using a separate config file per feature? a global locale/language file ?...)

claudiabcn (approved):
Approved!! Good changes. About the minor suggestion, I know that it would be out of scope, but we could think about it, how to manage that, for the next sprints.

kenan-it-academy (approved):

miguelm-montano (approved):

PR #693, opened 2026-05-27
Feature/593 ticketList refactor because of ticketRow

### LINKED ISSUE #593: [TICKETING] As a developer, I want to change the status of a ticket

claudiabcn:
In a previous PR, I created TicketRow which is a new standalone component that handles row rendering with a priority dropdown (admin/superadmin only) and static status display. 
Now, ticketList has been refactored to delegate row rendering to TicketRow, removing inline logic and simplifying the component. Tests have been updated also. TicketRow test file was previously added too.

**Please, this is de the last PR. It is very important to show the dropdown.**

image
image


triflip (approved):
Nice refactor!
TicketList is much cleaner now and the extraction of TicketRow makes the structure easier to maintain.
Everything looks good to me.

lxcxsito (approved):
Looks good for me

kenan-it-academy (approved):

PR #696, opened 2026-05-27
feat(#645): add date validation - end_date cannot be earlier than sta…

### LINKED ISSUE #645: Add start and end date fields to the project creation form

triflip:
Adds validation to prevent end_date from being earlier than start_date.

- Add validation in validateForm() to check end_date >= start_date
- Add test to verify valid dates pass validation without error

Closes #645


miguelm-montano (approved):
Minimal change!
It does what it says and closes the job completely.
Nice work Toni! ✅

H3llynx (changes requested):
Short and clear, but I’m not sure the PR will pass without adding the corresponding test in `FormCreate.test.tsx`, since the PR description mentions it should be included.

Also, the `end_date` field is auto-calculated, and users currently can’t edit it manually. If an error occurs, it’s therefore not something they can fix themselves. I’d suggest adjusting the error message to be more generic and less frustrating for the user (unless the idea is to make that field editable in the future).

triflip:
Hi Helen! Thanks for the feedback!
Regarding the test — there is a test included: should not show date validation error when end_date is after start_date. Since end_date is auto-calculated, the validation is defensive code and the test verifies valid dates don't trigger the error.
Regarding the error message — great point! Since the user can't edit end_date manually, I've updated the message to: "Hi ha un error amb les dates. Torna a seleccionar la data d'inici." — more actionable for the user.

miguelm-montano (approved):
Good call, @H3llynx !

Toni, thanks for the changes and given the context regarding the tests, everything looks good just as described

H3llynx (approved):
Fair enough about the test. Great job!

kenan-it-academy (changes requested):
Missing test (@H3llynx  @miguelm-montano seriously??)

triflip:
@kenan-it-academy , The validation is defensive code — since end_date is auto-calculated from start_date + duration, it can never actually be earlier than start_date through normal UI interaction. That's why the test verifies the opposite: that valid dates don't trigger the error.
However, I understand your concern. I can add a test using vi.spyOn to mock computeEndDate and force the invalid case.

H3llynx (approved):

miguelm-montano (approved):

kenan-it-academy (approved):
I think just closing this PR because the use case is essentially almost impossible was cleaner, but this is fine 😉

PR #771, opened 2026-06-04
Feature/725 ticketing make comment form editable

### LINKED ISSUE #763: Allow editing existing comment inline
PR 3 — Allow editing existing comment inline feat: make TicketCommentForm editable

## Context 
- The TicketCommentForm component currently renders a readOnly textarea when an existing comment is passed via initialValue, and hides the "Desar" button. This prevents users from editing their own comments. This PR updates the form component to dynamically allow editing only if the logged-in user is the author of the comment.
- The component remains independent of the API layer — it receives the user data and behavior via props.

## Criteria
- The textarea is editable if the comment belongs to the logged-in user (authorId === currentUserId)
- If the comment belongs to another user (authorId !== currentUserId), the textarea remains readOnly and uses a disabled background style
- The "Desar" button is visible when creating a new comment or editing your own comment, but hidden when viewing another user's comment
- Submitting calls onSubmit with the updated text
- Unit tests cover the happy path for editing own comments

## To Do
- Add editText local state initialized from initialValue
- Remove readOnly from the textarea when displaying an existing comment
- Update handleSubmit to use the correct text depending on mode (edit vs. new)
- Always show the "Desar" button
- Add happy path test

francobrida:
### Context
To implement comment editing simple and clean, we decided to make the existing comment textarea editable directly in the UI, rather than adding a separate "Edit" button mode. 
To prevent users from editing other teammates' comments, we check comment ownership (authorId === currentUserId). If the comment doesn't belong to the logged-in user, the textarea remains readOnly with a disabled visual style, and the submit button is hidden.

### Summary
- Added currentUserId prop and isOwner check to TicketCommentForm.tsx.
- Made the initialValue textarea editable and showed the "Desar" button only if isOwner is true.
- Retained the readOnly state and disabled background styling for comments belonging to other users.
- Added a happy path unit test in TicketCommentForm.test.tsx verifying the form submission with edited text for the owner.

Note for Reviewers / How to Test:
This PR only introduces the isolated UI component changes (PR 3 of 4). The page integration and hook connection will be done in the final sub-task (next PR). Therefore, this change cannot be tested in the browser yet (the comment will still appear as read-only on the page). Please verify this PR exclusively by running the unit tests.

### DIFF
--- a/frontend/src/components/ticketing/TicketCommentForm.tsx
+++ b/frontend/src/components/ticketing/TicketCommentForm.tsx
@@ -24,16 +24,23 @@ const TicketCommentForm = ({
 
   const isOwner = authorId === currentUserId;
 
+  const handleEditComment = async () => {
+    if (!editText.trim()) return;
+    await onSubmit(editText);
+  };
+
+  const handleNewComment = async () => {
+    if (!comment.trim()) return;
+    await onSubmit(comment);
+    setComment("");
+  };
+
   const handleSubmit = async (e: React.FormEvent) => {
     e.preventDefault();
-
     if (initialValue !== undefined) {
-      if (!editText.trim()) return;
-      await onSubmit(editText);
+      await handleEditComment();
     } else {
-      if (!comment.trim()) return;
-      await onSubmit(comment);
-      setComment("");
+      await handleNewComment();
     }
   };
 
--- a/frontend/src/components/ticketing/TicketCommentForm.tsx
+++ b/frontend/src/components/ticketing/TicketCommentForm.tsx
@@ -26,10 +26,15 @@ const TicketCommentForm = ({
 
   const handleSubmit = async (e: React.FormEvent) => {
     e.preventDefault();
-    const text = initialValue ? editText : comment;
-    if (!text.trim()) return;
-    await onSubmit(text);
-    if (!initialValue) setComment("");
+
+    if (initialValue !== undefined) {
+      if (!editText.trim()) return;
+      await onSubmit(editText);
+    } else {
+      if (!comment.trim()) return;
+      await onSubmit(comment);
+      setComment("");
+    }
   };
 
   const textareaClass =
--- a/frontend/src/components/ticketing/TicketCommentForm.tsx
+++ b/frontend/src/components/ticketing/TicketCommentForm.tsx
@@ -7,6 +7,7 @@ type Props = {
   initialValue?: string;
   authorId?: number;
   date?: string;
+  currentUserId?: number;
 };
 
 const TicketCommentForm = ({
@@ -16,9 +17,12 @@ const TicketCommentForm = ({
   initialValue,
   authorId,
   date,
+  currentUserId,
 }: Props) => {
   const [comment, setComment] = useState("");
   const [editText, setEditText] = useState(initialValue ?? "");
+
+  const isOwner = authorId === currentUserId;
 
   const handleSubmit = async (e: React.FormEvent) => {
     e.preventDefault();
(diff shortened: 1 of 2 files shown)


miguelm-montano (changes requested):
Franco, the implementation is very well done!

- `isOwner` is simple and clear
- The test updates the name and adds `authorId/currentUserId`

Something I’m noticing while testing the implementation is that I can’t edit my own comment and I’ve seen that the current user is:

`currentUserId={123}`

Shouldn’t it pass the Id of the authenticated user? 

Please confirm this part and I’ll do the re-review!

francobrida:
Miguel! thanks for the review, I just added to the PR explanation this last part that addreses the testing.

"Note for Reviewers / How to Test:
This PR only introduces the isolated UI component changes (PR 3 of 4). The page integration and hook connection will be done in the final sub-task (next PR). Therefore, this change cannot be tested in the browser yet (the comment will still appear as read-only on the page). Please verify this PR exclusively by running the unit tests."

You are absolutely right, in the real app, we must pass the authenticated user's ID.
Just to clarify: the currentUserId={123} you saw is only inside the unit test file (TicketCommentForm.test.tsx) as a mock value to test the ownership logic.

In the actual application, wiring the form component to the authenticated user ID and the page state is the scope of my next sub.task. This PR only defines the UI component and its internal logic.

Once PR 3 is merged, we will wire the real authenticated user ID in PR 4 and you'll be able to test it in the browser! Let me know if that makes sense

miguelm-montano (approved):
Approved after the explanation!

Thanks for the comment, Franco! I needed that note in the context to understand it better.

IMHO, this is good! Straightforward and with a single responsibility ✅

josemanuelriu-cmd (approved):
The code is clean and looks good.
I ran the TicketCommandTest locally and it works correctly.
Go for it!

frontend/src/components/ticketing/TicketCommentForm.tsx
```diff
@@ -16,14 +17,19 @@ const TicketCommentForm = ({
...
 
   const handleSubmit = async (e: React.FormEvent) => {
     e.preventDefault();
-    if (!comment.trim()) return;
-    await onSubmit(comment);
-    setComment("");
+    const text = initialValue ? editText : comment;
+    if (!text.trim()) return;
+    await onSubmit(text);
+    if (!initialValue) setComment("");
```
kenan-it-academy: I don't know if I understand this setComment here, why are we clearing it if there was no initialValue? Also maybe the check should be initialValue vs empty string?

kenan-it-academy (changes requested):
The changes are a little hard to follow, I think readability should be improved. Maybe renaming things or extracting to a named function?

francobrida:
Thanks for the tip! I extracted the logic into two named functions — handleEditComment and handleNewComment — so handleSubmit just decides which one to call. Also used initialValue !== undefined for a more explicit check. 
Tests still pass ✅

miguelm-montano (approved):
After reading Kenan's review and your solution, it all makes sense. it's something new we've learned today. 

Tests still working and everything looks more cleaner. ✅

frontend/src/components/ticketing/TicketCommentForm.tsx
francobrida: Thnks! Applied suggested changes!

josemanuelriu-cmd (approved):
Although I'm used to complex code, it's true that the change makes everything much easier to read. I'll make sure to point out that if the code is difficult to read, it should be changed in the future.

kenan-it-academy (approved):
Much better 😉

PR #783, opened 2026-06-08
Feat/get user points history

### LINKED ISSUE #745: Get user points history
### **Context**

The points history table and model exist and entries are being saved when points are awarded. However, there is no endpoint to retrieve this data, which means the frontend **cannot display** the history modal to the authenticated user.                 

### **Criteria**

- A new endpoint `GET /ligas/history` returns the authenticated user's points history in chronological order                                                               
- Each entry includes date, points and reason
- The endpoint is protected and only returns data for the authenticated user

### **To Do**

- Add `GET /ligas/history` route to `api.php` protected by Sanctum auth                  
- Create history method in `LigaController` that returns the authenticated user's entries from `liga_point_histories `                                                   
- Add tests for the endpoint

miguelm-montano:
### **Summary**

- Adds `GET /ligas/history` endpoint that returns the authenticated user's points history in chronological order      
- Each entry includes `points`, `activity` and `created_at`                                                               
- Endpoint is protected by Sanctum. Only authenticated users can access it

### **Test plan**

- [x] Run `docker exec backend php artisan test src/tests/Feature/LigaTests/LigaGetHistoryTest.php` -  3 tests pass
- [x] Verify authenticated user receives only their own entries
- [x] Verify unauthenticated request returns 401

### DIFF
--- a/backend/src/app/Http/Controllers/LigaController.php
+++ b/backend/src/app/Http/Controllers/LigaController.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\Http\Controllers;
 
 use App\Models\Liga;
+use App\Models\LigaPointHistory;
 use App\Models\User;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
@@ -66,6 +67,15 @@ class LigaController extends Controller
             ]);
 
         return response()->json($entries);
+    }
+
+    public function history(): JsonResponse
+    {
+        $history = LigaPointHistory::where('user_id', auth()->id())
+            ->orderBy('created_at', 'asc')
+            ->get(['points', 'activity', 'created_at']);
+
+        return response()->json($history);
     }
 
     public function addPoints(Request $request, User $user): JsonResponse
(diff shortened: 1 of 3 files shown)


claudiabcn (approved):
LGTM 👍

The new GET /ligas/history endpoint is correctly implemented and properly protected with Sanctum. It correctly filters by auth()->id() and returns the expected fields in chronological order.

The feature is well covered by tests, including authentication, authorization, and user isolation. Approved.

francobrida (changes requested):
Code looks good!

One thing: the summary says results are returned in chronological order, but I don't see a test covering that. If the orderBy direction changes, the tests wouldn't catch it. Maybe add a test verifying the order of results? or is it out of scope?

miguelm-montano:
@francobrida You're absolutely right. Thanks for pointing that out!! 
The PR now includes a test to verify that

francobrida (approved):
Tested locally — all 4 tests pass.

The endpoint is clean and minimal: filters by auth()->id(), selects only the needed columns and orders chronologically. Good call adding the ordering test after my comment :)

The +50 lines are justified — route, controller method and its feature tests are a single deliverable.

Approved!

kenan-it-academy (approved):

claudiabcn (approved):

PR #787, opened 2026-06-08
Feature/render button by contributor

### LINKED ISSUE #770: [OPTIONAL BONUS] button only appears to non-project-creator members
Only students who have joined a project can choose to leave it.

Project creators/owners cannot leave their own project.

to be considered:
In the future, creators may be allowed to transfer ownership to another member before leaving the project.

clara-cdp:
# Summary

Render the Deixar projecte button only for students who are contributors of the project, excluding the project creator.

# What Changed

- Added contributor id and user_id to ApiProjectContributor.
- Passed the project owner id from CodeConnectDetails to ProjectTeam.
- Used the current logged-in user id to check whether the user is a project contributor.
- Hid the Deixar projecte button when the current user is the project creator.

# Out Of Scope

Calling the delete contributor API.
Removing the contributor from the database.
Updating the team UI after leaving.
Changing TeamRow or ProjectParticipant.
Adding or changing the confirmation modal.

# Test

From the frontend folder:
```
npx vitest run src/components/code-connect/projectTeam/__test__/P
```

>We prioritized this feature first -even though is was added as bonus- because letting the UI know whether the current user is allowed to see the "Deixar projecte" button makes the future delete logic safer and simpler: the button only appears for valid users, and the delete handler will already have the correct contributor record available instead of needing to fetch it in a separate logic.

### DIFF
--- a/frontend/src/components/code-connect/projectTeam/ProjectTeam.tsx
+++ b/frontend/src/components/code-connect/projectTeam/ProjectTeam.tsx
@@ -13,6 +13,7 @@ interface ProjectTeamProps {
   contributors?: ApiProjectContributor[];
   timeDuration?: string;
   projectId?: number;
+  projectOwnerId?: number;
 }
 
 function ProjectTeam({
@@ -21,6 +22,7 @@ function ProjectTeam({
   contributors = [],
   timeDuration,
   projectId,
+  projectOwnerId,
 }: ProjectTeamProps) {
   const { getTeamByRole } = useProjectContributors(contributors);
   const frontendData = getTeamByRole("frontend");
@@ -48,10 +50,14 @@ function ProjectTeam({
   const frontendOffset = 0;
   const backendOffset = frontendData.emptySlots;
   const currentUserId = getCurrentUserId();
+  const isProjectOwner = currentUserId === projectOwnerId;
 
   const currentUserContributor = contributors.find(
     (contributor) => contributor.user_id === currentUserId,
   );
+
+  const shouldShowLeaveButton =
+    Boolean(currentUserContributor) && !isProjectOwner;
 
   return (
     <div className="flex flex-col items-start border border-gray-500 text-black w-80 pt-7 pb-10 px-6 rounded-3xl max-h-[700px]">
@@ -122,7 +128,7 @@ function ProjectTeam({
         >
           Apuntar-me
         </ButtonComponent>
-        {currentUserContributor && (
+        {shouldShowLeaveButton && (
           <div className="w-full flex justify-center -mt-8">
             <ButtonComponent
               className="my-5 w-full"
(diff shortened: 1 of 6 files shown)


francobrida (changes requested):
Hi Clara! The logic and the tests look really good.
I just found one critical detail:

I was actually struggling with this exact same issue while working on my Ticketing tasks: the getCurrentUserId() function is currently broken in the real app. If you take a look, it tries to fetch "user" from localStorage, but since there were recent auth changes, the project doesn't store the user object there anymore (only the token). Because of this, it will always return null in production, meaning the app will think no one is logged in and the 'Deixar projecte' button never renders.

(Side note: The reason the tests are passing is because getCurrentUserId is mocked to return 7, which forces the test to work, but hides the bug when running the actual app. I tested it locally).

To fix this, we can just use the global context we already have in place:

```
import { useUserContext } from "../../../context/UserContext";
// ... inside the component:
const { user } = useUserContext();
const currentUserId = user?.id; // 
```

Keep in mind you'll also need to update ProjectTeam.test.tsx to mock useUserContext instead of getCurrentUserId. Once that's changed, it should be good to go! 

Captura de pantalla 2026-06-08 224429

clara-cdp:
Hi Franco, thanks for the tip!

You were right: `getCurrentUserId()` was not reliable anymore because it depends on a `user` object in localStorage, and the app now gets the authenticated user from `UserContext`.

I updated `ProjectTeam` to use `useUserContext()` instead, and changed the tests to mock `useUserContext` rather than `getCurrentUserId`.

I also confirmed we still need the backend to return `id` and `user_id` for each contributor: `useUserContext` tells us who the current user is, while `contributor.user_id` lets us match that user against the project contributors. The contributor `id` will be needed later for the delete request.
Screenshot 2026-06-09 091151

francobrida (approved):
tested and working. all tests pass and the issue was fixed. 
Approved!

josemanuelriu-cmd (approved):
It passes the tests correctly, and the approach to obtaining the data (current user ID and project creator ID) is appropriate.
From my perspective, everything is fine.

frontend/src/components/code-connect/projectTeam/ProjectTeam.tsx
```diff
@@ -115,6 +129,20 @@ function ProjectTeam({
         >
           Apuntar-me
         </ButtonComponent>
+        {shouldShowLeaveButton && (
+          <div className="w-full flex justify-center -mt-8">
+            <ButtonComponent
+              className="my-5 w-full"
+              type="button"
+              variant="secondary" //should change to discret
+              onClick={() => {
```
kenan-it-academy: Easier to make PRs independent if you make this an empty named function elsewhere that you call her.

kenan-it-academy (changes requested):

frontend/src/components/code-connect/projectTeam/ProjectTeam.tsx
clara-cdp: Ok. I think I got you. In a real-case scenario it shall open a confirmation modal, but in this case I can just add a mock empty handler and keep everything separate till the next PR.
thanks!

kenan-it-academy (approved):

clara-cdp:
Branch update made line 35 to be duplicated, also on line 31. There was no conflict just an automatic test failure.
I deleted line 31. 
that's all that changed.

image

francobrida (approved):

PR #791, opened 2026-06-09
Introduce a minimal global context to manage project join requests ac…

### LINKED ISSUE #739: User pending request on listProjects and Project Detail
Create a global React Context that manages join request state across all Code Connect views (Project List + Project Detail).

This will replace local state management and ensure a single source of truth for join requests.

lxcxsito:
This PR introduces a minimal global context to manage project join requests.

It provides a shared state layer that allows synchronization of join actions across different views (Project Card and Project Detail).

## Features
- Add join request per project
- Prevent duplicate slot requests
- Global state accessible via React Context
- Unit tests for core logic

Future PRs will integrate this context into:
- Project Card view
- Project Detail view

### DIFF
--- a/frontend/src/context/ProjectJoinContext.tsx
+++ b/frontend/src/context/ProjectJoinContext.tsx
@@ -24,7 +24,7 @@ export function ProjectJoinProvider({ children }: { children: ReactNode }) {
       const current = prev[projectId] ?? [];
 
       const exists = current.some(
-        (s) => s.area === slot.area && s.index === slot.index
+        (s) => s.area === slot.area && s.index === slot.index,
       );
 
       if (exists) return prev;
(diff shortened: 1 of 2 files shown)


clara-cdp (changes requested):
Hi Lucas, 

So far is not passing the frontEnd tests... try to run the npx prettier

clara-cdp (approved):
Hi, 

Better now. Tests passed locally as well. 

LGTM!

gregoridelrio (approved):
Nice work, the context is clean and well-scoped.

One thing worth addressing before merge: there's an unresolved GitHub Actions warning on line 46:

> *Fast refresh only works when a file only exports components.*

Since the file exports both the `ProjectJoinProvider` component and the `useProjectJoinContext` hook (plus types), Vite's Fast Refresh can't track it properly. A common fix is splitting the hook into its own file or moving the types to `types/`.

Otherwise LGTM, tests cover the right cases and the PR is well scoped. 👍

PR #772, opened 2026-06-04
feat: add trigger button to leagues view for admin, superadmin and mentor

DanielVeraMunoz:
Summary

  - Adds a trigger button to the Leagues view to manually trigger the weekly league transition
  - Button is only visible to users with admin, superadmin and mentor roles
  - Rotate icon .svg format downloaded from Figma and added to src/assets/rotate.svg
  - Updated LeaguesPage tests to mock useUser hook and added tests for button visibility by role

Captura de pantalla 2026-06-04 a las 12 30 06

PR #800, opened 2026-06-10
Feature/connect points history endpoint

### LINKED ISSUE #757: Add points history table to modal
DESCRIPTION
As a user, I want to see a table inside the points history modal in order to review my points history with all relevant details.
OUT OF SCOPE

Layout, CSS
API integration / real data

ACCEPTANCE CRITERIA

The modal contains a table with three columns: Fecha, Puntos ganados and Actividades realizadas
The table renders correctly with mock or empty data

Image

miguelm-montano:
### **Summary**            
                                                                                           
- Added `useGetPointsHistory` hook that fetches the authenticated user's points history from `GET /api/ligas/history`                                                                                            
- Connected the hook to `LeaguesPage.tsx` The modal now loads real data from the backend
- **Updated** `PointsHistoryTable` to accept data as prop and render rows from the API response                     

### **Test plan**
- [x] Log in with GitHub and set your role to Student from the profile section then run `docker exec backend php artisan db:seed --class=LigaPointHistorySeeder` to populate history data
- [x] Open the "Veure el meu historial" modal. The table should display the seeded history entries with date, points and activity columns
- [x] `useGetPointsHistory` unit tests pass: `npx vitest run src/hooks/__tests__/useGetPointsHistory.test.ts`
- [x] `PointsHistoryTable` unit tests pass: npx vitest run `src/components/leagues-ranking/PointsHistoryTable/PointsHistoryTable.test.tsx`

image


orlandorojasoyarzun (changes requested):
Hey @miguelm-montano !
Thanks for the note about the duplicated code !, that clarifies a lot.

I found one bug that needs to be fix: 
- Token key mismatch (line 61 in endPointLeagues.ts),
- fetchPointsHistory uses "auth_token" (yours is the correct) line 22,
- triggerWeeklyTransition uses "authToken" (incorrect from before I guess). 
- This will cause the API call to fail with 401 bc it'll return null all the time

and, just to be sure .. the "Trigger" button is intentionally only a placeholder right? is not meant to be implemented the api call yet? 
- fx triggerWeeklyTransition() exists in endPointLeagues.ts but it's not being called anywhere, right?
- if this the case, you could maybe consider adding a comment, just for easy review.

Overall good !
I'm only scared about the amount of lines for the scope, but since you like to duplicate the code so much, 
it's well justified in the PR description ,) nice !

miguelm-montano:
@orlandorojasoyarzun Orlando thanks for pointing out all the issues, you're absolutely right

The token bug was already fixed in this branch with the latest commit, once this PR is merged it will be corrected in develop as well.

The 3 token related points all refer to the same issue and are now resolved here.

The "Trigger" button is there because branch #772 hasn't been merged yet. You can check it there, the button and the API call are already implemented and pass all the tests. Once it gets merged, this branch will reflect the updated state.

orlandorojasoyarzun:
Cool @miguelm-montano , I knew it was a decision for this pr  

And well done for fixing the bug ,)

orlandorojasoyarzun (approved):
approved !

josemanuelriu-cmd (changes requested):
In line 15 of the useGetPointsHistory.ts file, you're using `console.log`. For consistency with the project, I think it would be better to use `console.error`.

miguelm-montano:
Great catch @josemanuelriu-cmd You're absolutely right.
I've already updated the file to make sure everything is consistent. Thank you!!

josemanuelriu-cmd (approved):
With the requested change made, I see that everything is correct. Great job!

kenan-it-academy (commented):
I understand I need to wait for #801 to merge to review this and that's wht this is so long

kenan-it-academy (approved):

DanielVeraMunoz (approved):
good job!

PR #808, opened 2026-06-15
feat: highlight projects and avatar where user is an accepted member

### LINKED ISSUE #807: Highlight projects where user is a member
DESCRIPTION

As a student, I want the projects where I'm an accepted member to be visually highlighted in the project list, so that I can quickly identify my active collaborations.
OUT OF SCOPE

Filtering or hiding non-highlighted projects
Changes to the "Els meus projectes" button behavior

ACCEPTANCE CRITERIA

 Projects where the authenticated user has an accepted contributor status are highlighted with a primary-colored border
 The user's name appears in bold within the participant list of the highlighted project

gregoridelrio:
Description:

Highlights projects in the CodeConnect dashboard where the authenticated user is an accepted member. The project card gets a primary-colored border, the user's avatar gets a primary-colored outline, and the user's name is displayed in bold/black instead of gray.

Changes:

- Added isCurrentUser helper to ProjectCard to check if the authenticated user is an accepted participant
- Applied conditional border-primary to the card when the user is a member
- Applied conditional outline and rounded styles to the user's avatar
- Applied conditional text color to the user's name
- Added tests covering highlighted and non-highlighted cases

Figma:

image

Project:

image


orlandorojasoyarzun (approved):
Good work !

Some non blocker- suggestions:
- isCurrentUser could be a custom hook - Right now it's a local function that closes over user. Moving it to something like useProjectMembership would make it easier to reuse and test if the logic grows.
- Type safety - participant.user_id is optional (?number) and you're comparing directly with user?.id. If these types mismatch (string vs number, like: 1 === "1", i don't think it should happen tho but it would depend in how user.id is defined in the usercontext), the comparison could silently fails.

the PR is fine, well done !

clara-cdp (approved):
Nice, this is clear and nicely scoped.
Styling looks subtle and pretty and very well aligned with the global styles. 
Worked on local, it highlights and filters the projects as it should.

kenan-it-academy (approved):

PR #840, opened 2026-06-19
feat: add notification tracking fields to Ligas table and update model attributes

francobrida:
## What this PR does
Adds two new columns to the `ligas` table to support league-change notification tracking:
- `previous_league_id` (unsignedBigInteger, nullable): stores the user's previous league before a promotion/demotion.
- `notification_dismissed` (boolean, default false): tracks whether the user has acknowledged the notification.
Also updates the `Liga` model to reflect these fields in `$fillable`, `$attributes` and `$casts`.

## How to test
Run inside the container:
```bash
docker compose exec backend php artisan test src/tests/Feature/LigaTests/

### DIFF
--- a/backend/src/app/Models/Liga.php
+++ b/backend/src/app/Models/Liga.php
@@ -21,18 +21,23 @@ class Liga extends Model
         'status',
         'language',
         'league_id',
+        'previous_league_id',
+        'notification_dismissed',
     ];
 
     protected $attributes = [
-        'points_weekly'=> 0,
-        'status'=> 'Junior Coder',
+        'points_weekly'          => 0,
+        'status'                 => 'Junior Coder',
+        'notification_dismissed' => false,
     ];
 
     protected $casts = [
-        'points'=> 'integer',
-        'points_weekly'=> 'integer',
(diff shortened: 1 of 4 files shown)


clara-cdp (approved):
Hi Franco,
LGTM, fields have been added correctly and as described in the PR.
Tests passed locally too.
Don´t forget to update factory and seeders later as well.
Approved

gregoridelrio (changes requested):
Migration and model changes are clean and well-scoped.

One minor thing worth considering: `previous_league_id` is cast as `'integer'`, but since `league_id` uses `LeagueTypeEnum::class`, it might be worth casting `previous_league_id` to the same Enum for consistency.

Once that's addressed, happy to approve!

gregoridelrio (approved):
Perfect, thanks for the quick fixes! Approving now 🙌

francobrida:
Closing this PR. After discussing with Kenan, we agreed on an alternative 
persistence approach: instead of adding columns to the ligas table, we will 
create a dedicated league_weekly_results table to track transitions. 
This ensures idempotency and better separation of concerns.

PR #845, opened 2026-06-22
Added new fields to list_projects table

### LINKED ISSUE #817: Add new fields for complete projects
## Description:
Fields github_url + youtube_url | Necessary for the modal https://github.com/IT-Academy-BCN/ita-wiki-monorepo/issues/820
- github_url (string, nullable, max: 500 chars)
- youtube_url (string, nullable, max: 500 chars)

josemanuelriu-cmd:
### **Description**
This PR has been designed to create and save the data for the new fields (github-url and youtube_url) that will be used when a project is finished.

### **Context:**
Currently, there is no difference between an ongoing project and a completed one. With this pull request, we want to ensure that upon project completion, the URLs for both the project repository and a YouTube demo are included, which will be displayed when the project is closed.

### **Changes included:**
Added new URL fields (github_url and youtube_url) to list_projects table
Modified the ListProyectRequests for the two new fields
Updated the factory (ListProjectsFactory.php) and the seeder (ListProjectsSeeder.php) files for theses new fields too
Updated the fillable list with the 2 new fields

### DIFF
--- a/backend/src/app/Http/Requests/ListProjectRequest.php
+++ b/backend/src/app/Http/Requests/ListProjectRequest.php
@@ -35,7 +35,8 @@ class ListProjectRequest extends FormRequest
             'language_backend' => 'required|string|max:255',
             'language_frontend' => 'required|string|max:255',
             'roadmap' => 'nullable|array',
-
+            'github_url' => ['nullable', 'url'],
+            'youtube_url' => ['nullable', 'url'],
             'programming_role' => [
                 $this->isMethod('post') ? 'required' : 'nullable',
                 'string',
@@ -61,6 +62,8 @@ class ListProjectRequest extends FormRequest
             'language_frontend.required' => "The frontend language field is required.",
             'programming_role.required' => 'The programming role field is required.',
             'programming_role.in' => 'The programming role must be one of: Frontend Developer, Backend Developer, Fullstack Developer, Other.',
+            'github_url.url' => 'The GitHub URL must be a valid URL.',
+            'youtube_url.url' => 'The YouTube URL must be a valid URL.',
         ];
     }
 }
--- a/backend/src/app/Models/ListProjects.php
+++ b/backend/src/app/Models/ListProjects.php
@@ -30,8 +30,9 @@ class ListProjects extends Model
         'language_backend',
         'language_frontend',
         'description',
-        'roadmap'
-    
+        'roadmap',
+        'github_url',
+        'youtube_url',
     ];
 
     public function contributorListProject()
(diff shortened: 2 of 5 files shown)


orlandorojasoyarzun (changes requested):
Nicely approached but these URL fields will need validation at the model/request level to ensure they actually contain valid URLs (Laravel's url rule or a regex). Otherwise, any string could be saved there.

In ListProjectRequest.php :
Add to rules():
'github_url'  => ['nullable', 'url'],
'youtube_url' => ['nullable', 'url'],
And to messages():
'github_url.url'  => 'The GitHub URL must be a valid URL.',
'youtube_url.url' => 'The YouTube URL must be a valid URL.',

This is temptative tho, you could use another message

Without this, any string can be saved as a URL. We need validation to ensure data integrity.

clara-cdp (approved):
Hi, 

If the PR scope consists ONLY on adding these two fields to the list_projects table, then it LGTM: the migration correctly adds the github_url and youtube_url columns.

Orlando is correct on pointing that URL validation should be added, but that belongs in the request/model rather than in the migration itself. 
With that, said, do not forget to update the seeder/factory ;)

PR is  nice and short, but description could be a bit more detailed.

orlandorojasoyarzun:
good catch with the seeders @clara-cdp !

josemanuelriu-cmd:
made the changes indicated by colleagues

orlandorojasoyarzun (approved):
Feedback was taken good !

orlandorojasoyarzun (changes requested):
The Model needs to be updated

josemanuelriu-cmd:
Thanks, Orlando, for noticing. We did talk about it, but I'd forgotten.

orlandorojasoyarzun:
happened exactly the same to me regarding to my db migration 😅 @josemanuelriu-cmd

clara-cdp:
SInce you're implementing this broadly, I think you should indeed update the model and controller, not just the request.  I am guessing some test should be updated properly... ?

orlandorojasoyarzun (changes requested):
@josemanuelriu-cmd If you could update with a proper description the PR, well described with what you are intending with, it because It feels this PR could lead into misunderstanding for not being well defined of what this really is.

It would be super helpful ! and for the approval process

orlandorojasoyarzun (approved):
Nice !

josemanuelriu-cmd:
> SInce you're implementing this broadly, I think you should indeed update the model and controller, not just the request. I am guessing some test should be updated properly... ?

Hi Clara, in the main chat I specifically asked if tests were necessary when only adding fields to the database. Kenan replied that they weren't, that they're only necessary when modifying behavior. That's why I haven't added the tests.
Regarding modifying the controller, it's not necessary since both `store` and `update` use `$request` and the fields are already specified there.

kenan-it-academy (approved):

Issue #819, opened 2026-06-18
backend's new Endpoint to be able to "mark as complete" codeconnect projects


orlandorojasoyarzun:
## Description:
Create endpoint: PATCH /api/codeconnect/{id}/complete

## Validation Rules:
1. User must be authenticated
2. User must be the project owner (user_id matches)
3. The project must be in open status
4. If github_url provided, must be valid URL format
5. If youtube_url provided, must be valid URL format
6. Project status must change to "completed"

PR #848, opened 2026-06-22
Add new field in the data base to be able to mark as complete projects in future api call

### LINKED ISSUE #818: Add new field in the data base to be able to mark as complete projects in future api call
## Context

This PR is a must when it comes to be able to mark as complete a project. This will be an API call linked to the data field that it's been added to the database. This will allow to connect the button marked as complete and make the API call to bring the data of which projects are complete or not. This is necessary to change the whole 'real' status of a project, making them complete or in progress. 

Because as it is right now, the mark as complete button it's only a UI implementation.

## Description:
It has to be added the correct fields in the database to be able to make an api call for change status project's 

## Technical data
Enum status (in_progress/completed) | 
Migrations for the right status field, necessary for the endpoint https://github.com/IT-Academy-BCN/ita-wiki-monorepo/issues/819
Model's new field added
Backend test with a little twist (check the Note)

- status enum: in_progress, completed,
              default: in_progress

orlandorojasoyarzun:
## Description 
This PR adds a new "status" field to projects. Now we can know if a project is still **in progress** or already **completed**. 

This is going to be necessary for the api call at: https://github.com/IT-Academy-BCN/ita-wiki-monorepo/issues/819. This sets things up for the "Mark as complete" button to work properly and the app can return projects accordingly. 

The default value is "in progress"

### About the test
The test creates the model directly instead of using the factory because the factory generates a random status, which would make the test non-deterministic. We also use `refresh()` after creation because Eloquent doesn't load database defaults into the model automatically — without it, the test would fail since the status would be null instead of the default IN_PROGRESS.

### DIFF
--- a/backend/src/app/Enums/ProjectStatusEnum.php
+++ b/backend/src/app/Enums/ProjectStatusEnum.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Enums;
+
+enum ProjectStatusEnum: string
+{
+    case IN_PROGRESS = 'in_progress';
+    case COMPLETED = 'completed';
+    
+    public static function values(): array
+    {
+        return array_column(self::cases(), 'value');
+    }
+}
--- a/backend/src/app/Http/Requests/ListProjectRequest.php
+++ b/backend/src/app/Http/Requests/ListProjectRequest.php
@@ -5,6 +5,8 @@ declare (strict_types= 1);
 namespace App\Http\Requests;
 
 use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Validation\Rule;
+use App\Enums\ProjectStatusEnum;
 
 class ListProjectRequest extends FormRequest
 {
@@ -35,12 +37,12 @@ class ListProjectRequest extends FormRequest
             'language_backend' => 'required|string|max:255',
             'language_frontend' => 'required|string|max:255',
             'roadmap' => 'nullable|array',
-
             'programming_role' => [
                 $this->isMethod('post') ? 'required' : 'nullable',
                 'string',
                 'in:Frontend Developer,Backend Developer,Fullstack Developer,Other',
             ],
+            'status' => ['nullable', Rule::enum(ProjectStatusEnum::class)]
         ];
     }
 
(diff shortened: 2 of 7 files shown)


francobrida (changes requested):
Great start! The migration, the ProjectStatusEnum, and the validation rules in the Request look solid.

However, it looks like you missed updating the ListProject model itself?
I think is missing:

Add 'status' to the $fillable array so we can safely create or update it.
Add 'status' => ProjectStatusEnum::class to the $casts array so Laravel automatically converts the database string into an Enum instance when we access it.

Once the model is updated, I'll be happy to approve! :)

gregoridelrio (changes requested):
Good structure overall! The Enum is clean, the migration references it correctly, and it's great to see the factory, seeder, and request all updated in the same PR.

- Missing tests: it would be good to have at least a basic test verifying the column exists and that the default value is `in_progress`.

orlandorojasoyarzun:
I thought about it too @gregoridelrio It was actually on purpose to not have it but it's okay, I will add it. thanks !

gregoridelrio (approved):
All points addressed! Approving!

francobrida (approved):
Looks great now Orlando! The model updates and the test pass. Good catch on using refresh() for the defaults too. Approved! 👍

kenan-it-academy (changes requested):
seems strange to me that the whole PR is about marking as complete but there is no test case for the complete case 🤔

orlandorojasoyarzun:
Hi @kenan-it-academy, thanks for taking the time to review.

To give you more context on the PR scope:
- This PR is about **adding a new database field** (`status` enum with `in_progress` and `completed`), not the "mark as complete" logic itself.
- The actual "mark as complete" logic will be implemented in a separate PR linked to issue 
#819.
- The tests added here verify only what this PR changes: that the column exists and that the default value is `in_progress`.

I've updated the PR title to make the scope clearer:
"Add new field in the data base to be able to mark as complete projects in future api call"

Could you take another look with this context in mind? Thanks !

kenan-it-academy (approved):

josemanuelriu-cmd (approved):
The PR is exactly what we discussed. I even think running the test wouldn't have been necessary.
Good job.

PR #851, opened 2026-06-22
feat: add useLeagueNotification hook

### LINKED ISSUE #835: Connect the endpoint to the frontend
## Context

- The frontend needs an API function and a custom hook to call the dismiss-notification endpoint right after login and - expose the result to the UI. The hook should only run when the user is authenticated to avoid unnecessary requests.

## Criteria

- fetchLeagueNotification() function exists in endPointLeagues.ts and calls PATCH /ligas/dismiss-notification
- useLeagueNotification hook calls it on mount only when the user is authenticated
- Hook exposes { hasChange, direction, newLeagueId }
- Vitest test included

## To Do

- Add fetchLeagueNotification() to endPointLeagues.ts
-Create useLeagueNotification.ts hook
- Add Vitest test

francobrida:
## Update ( [26-06-2026] )
Following recent feedback, the architecture has been further simplified. We removed the localStorage tracking from the frontend. Now, the backend handles the time-window validation, making the frontend completely stateless.

## Summary
Adds the useLeagueNotification custom hook required to fetch and manage the user's weekly league transition status.

## Context
After team feedback, we simplified the architecture: instead of using a PATCH endpoint to persist the "dismissed" state in the database, the hook now:

- Calls GET /ligas/notification to check if the user changed leagues this week.
- Uses localStorage (keyed by year and week_number) to track whether the user already dismissed the notification, avoiding showing it more than once per week without any backend state.
- This keeps the backend completely stateless regarding the notification read state.

## Changes included
- Registered GET /ligas/notification endpoint in API config and endpoint layer.
- Added LeagueNotificationResponse discriminated union type (with year and week_number).
- Created useLeagueNotification custom hook to manage fetching, loading state, error handling, and localStorage-based dismissal.
- Updated strict happy-path unit tests using Vitest and React Testing Library, including a test that verifies the localStorage key is written on dismiss.

## How to test
Since the UI component is not yet injected into the layout (next PR), you can verify the hook functionality via unit tests:

```bash
npm test src/hooks/__tests__/useLeagueNotification.test.ts
````


orlandorojasoyarzun (approved):
Good! Implementation is solid and follows the project conventions.

Thorough PR description — good documentation practice.

Only to clarify: the `dismiss()` (in line 35-37 useLeagueNotification): only updates the local state (sets `{ hasChange: false }`) without calling the API to persist this. Is this intentional, or should it call `dismissLeagueNotification()` to persist the dismissal in the backend?

gregoridelrio (approved):
Good work! The discriminated union type is well designed, the AbortController cleanup is handled correctly, and the PR description is exemplary.
Approved!

francobrida:
Hi Orlando! Good catch.

Yes, this is intentional for now to keep this PR as small and atomic as possible (< 50 lines). Since we don't have a dedicated GET endpoint yet, we are firing the PATCH on mount to fetch and dismiss simultaneously in this initial iteration.

In the upcoming PRs we will separate the GET (fetch) and PATCH (dismiss on click) logic to prevent losing the notification on refresh. Merging this skeleton for now!

kenan-it-academy (changes requested):
I don't understand this entire PR. This seems very convoluted. Is the intention to just detect when I've gone up/down a league recently? Why is there a PATCH involved? 

This should just key off of existing backend information, we know when a change has happened. At most I could see maybe adding a "notified" field to the league promotions/demotions so it's only shown once, but for an initial version I'd even just say "we show this if it's been less than a day" or something super simple like that.

francobrida:
Hi Kenan! You were absolutely right, the PATCH was unnecessarily convoluted for an MVP.

I just pushed a new commit dropping the PATCH endpoint completely. The hook now reads the notification from a pure GET request and uses localStorage to check if the user has already dismissed it for that specific week.. hope this simplified approach works better

also changed the PR name and description to reflect it's new goal.

note.
The diff looks larger than the actual feature because ~40% are test additions (going from 1 to 3 test cases) and the rest reflects a full architectural refactor from PATCH+DB to GET+localStorage, based on team feedback from this same PR. The actual business logic fits in under 50 lines.

orlandorojasoyarzun (approved):

DanielVeraMunoz (approved):

kenan-it-academy (changes requested):
> Uses localStorage (keyed by year and week_number) to track whether the user already dismissed the notification, avoiding showing it more than once per week without any backend state.
> This keeps the backend completely stateless regarding the notification read state.

I think this is the opposite of what we want. We DON'T want localStorage data. We DO want, tracking state in the backend

francobrida:
@kenan-it-academy thanks for the feedback! I completely misunderstood your last review.

After re-reading your comment, I understood that the backend should be detecting league changes (using the league_weekly_results table and a time-based check), and the frontend doesn't need to track the dismissed state at all.

I've simplified the hook accordingly:

❌ Removed all localStorage read/write logic
✅ dismiss() now just clears the local React state
✅ The modal will reappear on page reload if the change is recent (which is acceptable for an MVP)

The backend endpoint (GET /ligas/notification) will use a 24h time window to determine if a change is "recent enough" to show — which keeps the frontend completely stateless.

clara-cdp (approved):
Hi Franco, 
Well, this was tough one...!
- The architecture is simpler now: the frontend just asks the backend “should I show this notification?” and does not need to own week-window logic or persistence.
- useLeagueNotification exposes notification, loading, error, and dismiss, which what  the component will need.
- The endpoint addition is nicely contained in the existing API/config pattern.
- The hook uses AbortController, so it is already thinking about fetch lifecycle cleanup.
- Tests cover the main happy path accordingly

nice work!

kenan-it-academy (approved):

Issue #833, opened 2026-06-19
Track league changes in the database


francobrida:
## Context
After a weekly transition, we need a way to track if the user has been notified about their promotion/demotion so we can show them a modal.

Thanks to the introduction of the league_weekly_results table (which already tracks from_league and to_league), we no longer need to store the previous_league_id in the ligas table. We only need a flag to track the notification state.

One new field is needed:

notification_dismissed — tracks whether the user has already seen and dismissed the modal for their latest transition (boolean, true by default so new users aren't spammed).

The promotion/demotion logic must reset notification_dismissed to false each time a user changes league, so that recurring promotions in future weeks always trigger a new modal.

## Criteria
-New column notification_dismissed (boolean, default true) added to ligas table.
-ProcessLeaguePromotions resets notification_dismissed to false only for users whose league changes.
-Users who do not change league are not modified.
-PHPUnit test verifies the flag is correctly set.

## To Do
-Create migration adding notification_dismissed to ligas.
-Update $fillable and $casts in Liga.php.
-Update ProcessLeaguePromotions to set notification_dismissed = false before updating the league.
-Add PHPUnit test.

francobrida:
Closing this sub-issue. After team review (Kenan's feedback), we pivoted away from storing the dismissed state in the database to keep the backend stateless.

The final approach uses:

The existing league_weekly_results table (Lauren's PR #850) to detect if a league change occurred.
localStorage on the frontend (keyed by year + week_number) to track if the user already dismissed the modal for that specific week.
This removes the need for the notification_dismissed column entirely.

PR #852, opened 2026-06-22
feat: add LeagueNotificationModal UI component

### LINKED ISSUE #836: Build the league notification modal
## Context

A modal is needed to inform users of their league change after the weekly transition. The same component handles both promotion and demotion with different content. There is no close button — the user must click "Got it" to dismiss it. This is intentional: if the user could close the modal without clicking the button, the backend flag would not be consumed and the modal would reappear on every subsequent login.

## Criteria

- LeagueNotificationModal accepts direction: 'up' | 'down' and onDismiss: () => void
- direction="up" shows: "Congratulations, you've been promoted!" + "Got it" button
- direction="down" shows: "Sorry, you've been demoted!" + "Got it" button
- No close button — only the "Got it" button triggers onDismiss
- Vitest test included for both directions

## To Do

- Create LeagueNotificationModal.tsx
- Add Vitest test for both direction values

francobrida:
## Description
Creates the LeagueNotificationModal UI component to display weekly league transition results to the user.

## Context: 
This PR corresponds to the visual representation of the notification (Child 4 of US #833). It is completely independent and acts as a "dumb component". It receives props from its parent and does not contain API fetching logic or global state context (which will be implemented in the final PR).

## Changes included:

- Created LeagueNotificationModal component reusing the project's modal visual layout.
- Decoupled from the base Modal.tsx to force user interaction (removed "X" close button and backdrop click dismissal).
- Translated copies to Catalan with a motivating tone.
- Added strict happy-path rendering test using Vitest and React Testing Library.

## How to test
Since the component is not yet injected into App.tsx (next PR), you can verify it locally by running its unit tests:

Checkout this branch: git checkout feature/833-4-notification-modal
Enter the frontend container or local environment.

Run the isolated test:
```bash
npm test src/components/Modal/LeagueNotificationModal.test.tsx
```
Verify that the test passes and strictly covers the successful rendering and UI interaction.

### DIFF
--- a/frontend/src/components/Modal/LeagueNotificationModal.test.tsx
+++ b/frontend/src/components/Modal/LeagueNotificationModal.test.tsx
@@ -1,0 +1,29 @@
+import { render, screen, fireEvent } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import { LeagueNotificationModal } from "./LeagueNotificationModal";
+
+describe("LeagueNotificationModal", () => {
+  it("renders correctly and calls onDismiss on click", () => {
+    const handleDismiss = vi.fn();
+    render(
+      <LeagueNotificationModal
+        direction="up"
+        newLeagueId={2}
+        onDismiss={handleDismiss}
+      />,
+    );
+
+    // Verifica que se renderiza con el texto esperado en catalán
+    expect(screen.getByText("Felicitats!")).toBeInTheDocument();
+    expect(screen.getByText("Silver")).toBeInTheDocument();
+    expect(
+      screen.getByText(/El teu esforç ha donat fruits!/i),
+    ).toBeInTheDocument();
+
+    // Verifica la interacción
+    const button = screen.getByText("D'acord");
+    fireEvent.click(button);
+
+    expect(handleDismiss).toHaveBeenCalledTimes(1);
+  });
+});
--- a/frontend/src/components/Modal/LeagueNotificationModal.test.tsx
+++ b/frontend/src/components/Modal/LeagueNotificationModal.test.tsx
@@ -13,14 +13,12 @@ describe("LeagueNotificationModal", () => {
       />,
     );
 
-    // Verifica que se renderiza con el texto esperado en catalán
     expect(screen.getByText("Felicitats!")).toBeInTheDocument();
     expect(screen.getByText("Silver")).toBeInTheDocument();
     expect(
       screen.getByText(/El teu esforç ha donat fruits!/i),
     ).toBeInTheDocument();
 
-    // Verifica la interacción
     const button = screen.getByText("D'acord");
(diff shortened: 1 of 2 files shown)


josemanuelriu-cmd (commented):
Hi Franco, just one quick question.
In the LeagueNotificationModal.tsx file, you create the content on line 25 using useMemo(). I don't quite see the purpose of this. Wouldn't a simple isUp conditional be sufficient, returning one value in one case and another in the other? Please clarify so I can approve or reject it.

francobrida:
Thanks Riu! You are absolutely right. I added useMemo out of habit to follow our general optimization guidelines, but for a simple stateless component like this, it's definitely an unnecessary micro-optimization. I completely agree with keeping it simple. I pushed a commit removing it right away. 
Thanks for the catch!

josemanuelriu-cmd (approved):
Now it's LGTM

josep100:
El componente me parece correcto y cumple con el objetivo descrito en la PR.

Como posible mejora de mantenibilidad, he visto que dentro del if/else hay bastante JSX duplicado entre los estados up y down. Quizá podrías valorar extraer únicamente los elementos dinámicos (mensaje e icono) a constantes y reutilizar una única estructura JSX para ambos casos.

No lo considero bloqueante para la aprobación, simplemente una sugerencia de refactorización para reducir duplicación y facilitar futuros cambios.

josep100 (approved):

francobrida:
Thanks for the review and the great suggestion, @josep100 !

You're absolutely right about the DRY principle. I've just pushed a new commit extracting the dynamic elements (the icon and the message) into variables, so we are now returning a single JSX structure for both states.

frontend/src/components/Modal/LeagueNotificationModal.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { describe, expect, it, vi } from "vitest";
+import { LeagueNotificationModal } from "./LeagueNotificationModal";
+
+describe("LeagueNotificationModal", () => {
+  it("renders correctly and calls onDismiss on click", () => {
+    const handleDismiss = vi.fn();
+    render(
+      <LeagueNotificationModal
+        direction="up"
+        newLeagueId={2}
```
kenan-it-academy, line 11: Unrelated, but I'm worried about these IDs being used in reference to Named Leagues (like Silver in this case). This seems like a potential source of errors.

kenan-it-academy (approved):
Code looks goods

frontend/src/components/Modal/LeagueNotificationModal.test.tsx
francobrida: Got it. I'll refactor this to pass the name directly instead of the ID in the next PR when we inject the modal. Merging this for now

PR #850, opened 2026-06-22
feat: create league_weekly_results table and tests

gregoridelrio:
Description:

I want a league_weekly_results table to log weekly league transitions, so that the promotion/demotion process is idempotent and can run multiple times without side effects.

Out of scope:

- Frontend changes
- Notification logic

Acceptance Criteria:

- Migration creates league_weekly_results table with user_id, week_date, from_league, to_league
- Unique constraint on (user_id, week_date) ensures idempotency
- Tests included

PR #853, opened 2026-06-23
feat: add notification_dismissed column to ligas table

### LINKED ISSUE #833: Track league changes in the database

francobrida:
## Description
Creates the necessary database changes to support the League Promotion notifications. As part of resolving #833, this PR implements the database migration to store the notification state of a user.

**Architectural Note:** Although tracking this state could theoretically belong in the newly proposed league_weekly_results table, we decided to implement the notification_dismissed flag directly in the ligas table. This decision was made to avoid blocking our workflow, as depending on the league_weekly_results table would require waiting for PR #850 to be merged into develop first (complying with the team's "no branching off branches" rule). The refactor to move it in the future, if deemed necessary by the Lead, is extremely low-effort.

## Changes included:

-Created migration to add notification_dismissed boolean column to the ligas table.
-Default value is true so new users aren't immediately spammed with a fake notification.
-Updated Liga model $fillable, $attributes, and $casts to properly support the new column natively.
-Updated LigaFactory to include the new field.
-Added PHPUnit tests to LigaTableTest to assert the column exists and properly defaults to true when created via Eloquent.

## How to test
Checkout feature/833-a-notification-flag

Run php artisan migrate inside the container.

Run the complete test suite to ensure the new model attributes haven't broken any existing database interactions or factories:
```bash
php artisan test
```
Verify all tests pass globally and the database has the new column.

### DIFF
--- a/backend/src/tests/Feature/LigaTests/LigaTableTest.php
+++ b/backend/src/tests/Feature/LigaTests/LigaTableTest.php
@@ -21,6 +21,7 @@ class LigaTableTest extends TestCase
         $this->assertTrue(Schema::hasColumn('ligas', 'status'));
         $this->assertTrue(Schema::hasColumn('ligas', 'language'));
         $this->assertTrue(Schema::hasColumn('ligas', 'league_id'));
+        $this->assertTrue(Schema::hasColumn('ligas', 'notification_dismissed'));
     }
 
     public function test_points_weekly_defaults_to_zero(): void
@@ -74,6 +75,20 @@ class LigaTableTest extends TestCase
             ? $liga->status->value
             : $liga->status;
         $this->assertEquals('Junior Coder', $statusValue);
+    }
+
+    public function test_notification_dismissed_defaults_to_true(): void
+    {
+        $user = User::factory()->create();
+
+        $liga = Liga::create([
+            'user_id' => $user->id,
+            'points'  => 0,
+        ]);
+
+        $liga->refresh();
+
+        $this->assertTrue($liga->notification_dismissed);
     }
 
     public function test_existing_fields_are_not_affected(): void
--- a/backend/src/app/Models/Liga.php
+++ b/backend/src/app/Models/Liga.php
@@ -21,18 +21,21 @@ class Liga extends Model
         'status',
         'language',
         'league_id',
+        'notification_dismissed',
     ];
 
     protected $attributes = [
-        'points_weekly'=> 0,
-        'status'=> 'Junior Coder',
+        'points_weekly' => 0,
+        'status' => 'Junior Coder',
+        'notification_dismissed' => true,
     ];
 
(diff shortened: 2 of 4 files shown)


gregoridelrio (approved):
Clean and well-justified PR! The decision to default `notification_dismissed` to `true` makes sense — no need to show a modal to new users on registration. The architectural note explaining why it lives in `ligas` instead of `league_weekly_results` is a nice touch too. Model, factory and tests all covered. Approving!

DanielVeraMunoz (approved):
Nice work! Code is clean and well structured.
Good call on the default true to avoid spamming existing users.

backend/src/tests/Feature/LigaTests/LigaTableTest.php
```diff
@@ -76,6 +77,20 @@ public function test_status_defaults_to_junior_coder(): void
         $this->assertEquals('Junior Coder', $statusValue);
     }
 
+    public function test_notification_dismissed_defaults_to_true(): void
```
kenan-it-academy, line 80: we should default to false, right? It shows to them and will be false until the user dismisses it?

kenan-it-academy (changes requested):

backend/src/tests/Feature/LigaTests/LigaTableTest.php
francobrida: Good point! We initially defaulted to true to avoid spamming all existing users on deployment. However, since Lauren's PR #850 with the league_weekly_results table was just approved, we realized we don't even need this DB column or the PATCH endpoint anymore..

We are pivoting to key off of his table for the transitions, and we will use localStorage on the frontend to track the dismissal per week. I'll go ahead and close this PR to keep the backend stateless. Thanks for the review!

PR #858, opened 2026-06-23
When ticket status change to closed user who made the ticket gets 15 …

### LINKED ISSUE #843: [TICKETING] Award points to student after Bug Bounty resolution UI
As a student, I want to automatically receive the configured amount of points when my Bug Bounty is resolved, so that my league progression reflects my achievements.

lxcxsito:
When ticket status change to closed user who made the ticket gets 15 points 
#Modifications:
TicketController -> updateStatus checks if new status is closed and calls ligaController function -> addPoints
Test implementation


gregoridelrio (changes requested):
Hi! The implementation works correctly and it's great that you included tests. 👍

A small thing to do:
- The value 15 is hardcoded in multiple places — worth extracting to a constant.

When its changed, i will be happy to approve it!
Thanks for the work! 🙌

lxcxsito:
> Hi! The implementation works correctly and it's great that you included tests. 👍
> 
> A small thing to do:
> 
> * The value 15 is hardcoded in multiple places — worth extracting to a constant.
> 
> When its changed, i will be happy to approve it! Thanks for the work! 🙌

Luis told it's always 15 points

clara-cdp (changes requested):
Hi,
I completely agree with Lauren here. As you pointed out (no pun intended), the point value is always 15. Creating a points constant will not only make the code cleaner but will also facilitate any future changes. For instance, if the assigned points change to 5 later on, we would only have to update the constant value rather than tracking down every place it was hardcoded.

Once this is updated, the rest looks good to go!

clara-cdp (changes requested):
Hi Lucas,

The code does look much better with the variable implementation. Although, I'd still go for that constant to be used globally rather than a variable inside an if-else statement, and it should be used in the test as well so we don't have to update the test every time the value changes.    
It's a good practice to replace hardcoded integers ("magic numbers") with dedicated constants. The code will be much more robust and scalable, should you introduce more ways to reward points.

public const BUG_BOUNTY_REWARD_POINTS = 15; // or similar

clara-cdp (approved):
Awesomo! 

approved!

gregoridelrio (approved):
Thanks for addressing the feedback. LGTM!

kenan-it-academy (approved):
This probably doesn't go in this PR (hence why it's approved) but we should probably discriminate between a positive bug closure and a negative bug closure (for example due to a misreporting). I'd add that to backlog

PR #861, opened 2026-06-24
Fix/change category from nullable to required

### LINKED ISSUE #828: show suggestions button on Student view
As a student I expect the possibility of seeing some else's tickets with "suggestion" category

clara-cdp:
# Update 'category '
### Description:
Updated the category field on the tickets table from nullable to required. This ensures filtering tickets properly and ias it will be necessary when implementing the upvoting system in the future.
### what's changed
- Added a new migration:
    _2026_06_24_change_category_to_required_in_tickets_table.php_
- Updated tests where 'category' was absent or had to be modified 
- deleted "A user ca create a ticket without category test"

### Out of scope 

- Modifying any Requests such as UpdateTicketRequest or and there is no modifying ticket funtion on UI.
- Modifying TicketController as validation is already handled by CreateTicketRequest 

### Test
```docker compose exec backend php artisan test ``` 

image


gregoridelrio (approved):
Looks good! The migration safely backfills existing null categories before making the column required, good defensive approach. Approved.

lxcxsito (approved):
Looks good.

One thing to keep in mind for future PRs is keeping the scope as small as possible and limiting the number of touched files when feasible. In this case, the extra changes are understandable since the validation change required updating the affected tests.

kenan-it-academy (approved):

PR #860, opened 2026-06-24
feat(#855): add ticket detail modal with all fields

DanielVeraMunoz:
### Context
Second part of the ticket detail feature (see PR #855-A for the skeleton). This PR adds all the fields to the modal. The comment section is handled separately in PR #856 as it requires a different API call (`useTicketComments`).

### Description
Implements `TicketDetailModal` with all fields: ID, description, status, priority (with color), category (with icon), date (formatted) and role. Uses existing project assets (`Calendar.svg`, `user2.svg`), `CategoryIcon` component and constants (`STATUS_LABELS`, `PRIORITY_LABELS`, `PRIORITY_COLORS`, `CATEGORY_LABELS`) from `ticketConstants`. The component is purely presentational — no API calls inside.

> **Note on line count:** This PR is ~176 lines (component + test), above the usual target. This is a new component where most lines are Tailwind layout and field structure — there is only one responsibility: render ticket data already fetched by `TicketingPage`. Splitting it further would mean PRs with half a modal, which felt harder to review than a single self-contained component.

### Out of scope
- Comment section (PR #856)
- Wiring to TicketingPage (PR #855-A)

### Acceptance criteria
- Modal shows ID, description, status, priority with correct color, category with icon, formatted date and role
- Handles null priority and null category gracefully (shows "-")
- Test included (happy path)

Issue #855, opened 2026-06-23
Show ticket detail modal with all fields


DanielVeraMunoz:
**Description:**
As a user, I want to see all the relevant information of a ticket in a modal, so that I can read the details clearly.

**Out of scope:**
- Comment section
- Editing ticket information

**Acceptance criteria:**
- Clicking a ticket row opens `GenericModal`
- The modal shows: ID, descripció, estat, prioritat (amb color), categoria (amb icona), data, rol
- The modal can be closed
- Test: modal shows correct ticket data

Issue #854, opened 2026-06-23
Make ticket ID clickable to select a ticket


DanielVeraMunoz:
**Description:**

As a user, I want to click on the ticket ID so that the system knows which ticket I want to view in detail.

**Out of scope:**

- Modal content
- Comment section

**Acceptance criteria:**

- Clicking the ticket ID sets the selected ticket in state
- The ID has cursor and hover styles (consistent with project dropdowns) indicating it's clickable
- A tooltip "Veure detall" appears on hover
- Test: clicking the ID calls the handler with the correct ticket

PR #865, opened 2026-06-25
feat(#856): show ticket comment in detail modal

### LINKED ISSUE #856: Show comment in ticket detail modal
**Description:**
As a user, I want to see the comment of a ticket inside the detail modal, so that I have the full context without navigating elsewhere.

**Out of scope:**
- Comment form or editing
- Multiple comments

**Acceptance criteria:**
- If a comment exists, it appears as read-only text in the modal
- If no comment exists, shows "No hi ha comentaris" in muted gray
- Test: shows comment / shows fallback text

DanielVeraMunoz:
### Context
This is the last part of a 3-PR series implementing the ticket detail feature:
- **PR #854:** adds the clickable ID button with `onViewDetail` handler
- **PR #855-A:** wires up the modal skeleton in `TicketingPage`
- **PR #855-B (PR #860):** adds all ticket fields to the modal (in review)
- **PR #856 (this one):** adds the comment section inside the modal

### Description
Adds a read-only comment section to `TicketDetailModal`. Uses `useTicketComments` to fetch the comment for the current ticket. If a comment exists, it is displayed in a styled box; otherwise it shows "No hi ha comentaris". No editing — purely presentational.

This PR is independent from PR #860 (fields). When #860 merges, the conflict resolution will combine both: fields above, comment section below.

### Out of scope
- Comment editing (handled by the existing standalone `TicketCommentForm`)
- Ticket fields (PR #860)

### Acceptance criteria
- Modal shows the comment text when a comment exists
- Modal shows "No hi ha comentaris" when there are no comments
- Test included (happy path: with comment and without comment)

Grabación de pantalla 2026-06-25 a las 12 56 31


francobrida (approved):
Clean PR, exactly one responsibility. The tests cover both cases well.

One small thing to keep in mind for a future PR: calling useTicketComments(ticket?.id ?? 0) with 0 as fallback may trigger an unnecessary fetch when the modal is closed?. You could guard it with an early return or a conditional inside the hook maybe. Not a blocker for this MVP slice though.
Approved!

DanielVeraMunoz:
> Clean PR, exactly one responsibility. The tests cover both cases well.
> 
> One small thing to keep in mind for a future PR: calling useTicketComments(ticket?.id ?? 0) with 0 as fallback may trigger an unnecessary fetch when the modal is closed?. You could guard it with an early return or a conditional inside the hook maybe. Not a blocker for this MVP slice though. Approved!

Thanks! Good catch, actually the hook already guards against it: if (!ticketId) return treats 0 as falsy and skips the fetch, so no unnecessary requests are made.

clara-cdp (changes requested):
Ah, Dani, code is great and PR well resolved but now this is happening:
when you press on "comentari" you also get a dialogue box.    
I personally like your UI much better, but the other one also let´s you write and modify comments (Although you have to refresh the page as desar does not sync). 
I would move that logic to your modal or find another solution. 
Shame, as your work is great, but I think this should be targetted now, maybe let´s talk to Luis and Kenan? 
not sure if I should approved, sorry


Image

francobrida (changes requested):
Hi Dani and Clara! I can add context here since the comment dialogue you're seeing was part of my work from the previous sprint.

Clara is right — there's a UX conflict now that both exist simultaneously. The code in this PR is clean and isolated, but the interaction design needs to be resolved first. on my first review I assumed this would be addressed on a next PR.

I agree we should talk in Kenan and Luis before merging this, so we can align on whether to consolidate both into a single modal or keep them separate with different triggers.

clara-cdp:
I was redeading the PR description, am I right to understand that this will be inside the current modal, you've just build with the ticket detail? it that is the case, maybe there is no so much of a conflict, but just another way to see the comment... enlighten us @DanielVeraMunoz 
"Adds a read-only comment section to TicketDetailModal. "

DanielVeraMunoz:
Hey team! @clara-cdp @francobrida  Just to clarify the scope of this subtask: this modal lives inside the other section, and my subtask is specifically to display ticket property details in a read-only view. Making changes like updating the category, priority, or adding/editing comments is out of scope for me. The modal design has been approved by Luis.

The reason this is split into two PRs is that this one involves an API call, while the other PR doesn't interact with that API at all — plus combining them would have made for a very large diff. That said, everything is explained in the descriptions of both PRs.

Separately, there were discussions about potentially reworking the Ticketing page view — possibly showing fewer columns and allowing users to edit things like category, priority, or comments directly from that view. That could be a good place to address those interactions down the line. For now, this is purely a frontend implementation with no backend work in my scope.

Hope that clears things up!

image

francobrida (approved):
Thanks for the clarification, Dani! Scope is clear now and the modal looks great. The UX conflict with the existing comment section is something we'll need to address in a follow-up issue. Re-Approved!

clara-cdp (approved):
Thanks Dani, apologies for the hold-up. You had such a large amount of work separating everything into small PRs, which was indeed quite hard. Kudos for how well you organized it.

Approved!

kenan-it-academy (approved):

josemanuelriu-cmd (approved):
Hi Dani, from what I can see, everything is well-designed. It works according to the scope.
I did notice one small bug, though I don't think it falls within the current scope. It happens when you open the modal for a ticket that has no comments (so it appears empty), close the modal, create a comment, and then reopen the modal—it still shows that there are no comments. It can be resolved by opening a different modal or reloading the page.
I'm approving it anyway.

josemanuelriu-cmd (approved):
Hi Dani, with the latest change made, everything seems to be working perfectly.
Great job.

PR #867, opened 2026-06-25
Feature/show suggestions button if student

clara-cdp:
# "Veure suggeriments" button renders only if the user is a student. 

### what's changed
- added a small function that will render the "veure suggeriments" button only if the user is a student.
- added relevant tests

### out of scope
- filtering functionality.


orlandorojasoyarzun (changes requested):
Well done ! 

Just to suggest to use the project's roles constants instead of the string literal:

```tsx
import { roles } from "../data/tempRoles";

const isStudent = user?.role === roles.STUDENT;
```
instead of `const isStudent = user?.role === "student";`

This is the pattern already used in TicketRow.tsx for the admin/superadmin check. 

The button is empty as you specified in the description as out os scope right? you leaving the empty space for a future pr to be done in there?

orlandorojasoyarzun:
Tested locally in the app and works as it should ! 💪

clara-cdp:
Thanks Orlando, well spotted, here are the changes required.

orlandorojasoyarzun (approved):
Nice !

DanielVeraMunoz (approved):
Clean and minimal. isStudent reads well, good call using the roles constant instead of a hardcoded string. The test covers both cases properly. Approved!

kenan-it-academy (approved):

francobrida (approved):

PR #869, opened 2026-06-25
Expose status field in index and show methods - added doc accordingly

### LINKED ISSUE #868: Expose status field in API responses
## Context
The show and index methedos in the controller, now include a status field in their responses. This updates the existing tests to check
for this new field, so they pass again. The status value comes from the project itself since the factory generates it randomly (in_progress or completed).

orlandorojasoyarzun:
## Context
When I implemented #848 (add status field to the data base to be able to mark projects as complete), I forgot to expose the new field in the API responses. Without this change, the frontend cannot access the project status.

I'm addressing this oversight in a separate PR to keep the scope focused and easier to review.

## Changes
- `index()` endpoint now returns the `status` field for each project
- `show()` endpoint now returns the `status` field
- OpenAPI/Swagger documentation updated accordingly
- Some test has also been refactored as now they need the status 

## Related
- Follows up on 
#848

- Sets things up for the complete project endpoint 
#819


lxcxsito (approved):
Nothing to complain about. Approved!

francobrida (approved):
Hi Orlando! Lgtm!

kenan-it-academy (approved):

PR #870, opened 2026-06-25
status "complete" in codeconnect project's new endpoint

### LINKED ISSUE #819: backend's new Endpoint to be able to "mark as complete" codeconnect projects

orlandorojasoyarzun:
## Description
- This PR adds the backend endpoint to mark Codeconnect projects as "completed" 

NOTE: This is a very big PR, I'm not sure If I could had separated better, but I'm all ears for suggestions if needed

- This sets up the backend for the "Mark as completed" button feature  that will be added to the frontend 
- New projects will start by default `in_progress`
- Without this  endpoint, there's no way to persist the "completed" state in the  database.
- Documentation was added accordingly (many many lines)

**NOT APLICABLE IN THIS PR**: To mention because if you check on the commit historial, you could tell at some point this PR was in charge of doing this point, then, It was split into an independent one and it's already merged:
- Exposes the `status` field in the existing `index()` and `show()` at `ListProjectController`, so the frontend can know whether a project is in progress  or completed

## Note about tests
For now, only the happy path test is included (verifying that an owner can successfully mark their project as completed). Additional tests covering error cases (unauthorized users, invalid URLs, already-completed projects, etc.) can be added later if needed. The endpoint been tested in Postman accordingly

## Demo
Here is a little demo: 
https://github.com/user-attachments/assets/6700dd0c-cf68-4e03-8cf8-d6b8f489875a


josemanuelriu-cmd (changes requested):
Hi Orlando,

I see a couple of things that I think should be modified before this can be approved:
- In `api.php`, you are setting the new endpoint route using `.../{id}/...`. If you look closely, the other routes use `.../{listProject}/...`. For the sake of consistency, I would implement it the same way as the other routes.
- I noticed the message `'message' => 'Project is already completed'` on line 443 of `ListProjectsController`. You are assuming there will always be only two states in the enum—which is true right now—but if we were to add more in the future, this message would no longer be accurate. For better consistency, I would change the message to "Project must be in 'in_progress' status to mark it as completed."

orlandorojasoyarzun:
Changes done @josemanuelriu-cmd 💪

francobrida (changes requested):
Hey Orlando! Great work on this PR! I ran the test (CompleteProjectTest) inside the Docker container and everything is passing and working good.

I just have two minor, non-blocking suggestions for consistency:

Route parameter consistency: In api.php, the parameter is defined as {listProject}. 
In the controller's complete method, it is received as $id. 
Since Laravel maps route parameters positionally, this works fine, but changing the controller argument to $listProjectId (or $listProject) would make it cleaner and consistent with the other routes.

Swagger Docs: In the OpenAPI annotations for the complete method, you could update {id} to {listProject} (both in the path and the @OA\Parameter name) to match the actual route definition.

Other than that, it looks awesome. Let me know when you check these out and I'ts good to go!

gregoridelrio (changes requested):
Hi Orlando! Agree with the suggestions above, the route parameter consistency is a good fix, and don't forget to update the {id} in the OpenAPI annotations to {listProject} as well to keep everything consistent.

The logic is solid and the docs are thorough. I'll approve once those changes are in!

orlandorojasoyarzun:
@francobrida y @gregoridelrio Cool ! 
I completely forgot, thanks ! 
it's FIxed !

francobrida (approved):
with those corrections now Lgtm! approved!

gregoridelrio (approved):
Thanks for the changes! Approved!

kenan-it-academy (changes requested):
The main issue with this PR is it handles too many use-cases for a single PR. Notice how many of the possible flows in the `complete()` function aren't even tested for. Those flows (and their tests and documentation) should each be a separate PR (or at least have 2-3 groups of related functionalities)

orlandorojasoyarzun:
NECESITA TEST Y necesita estar agrupados los casos acordes en los test 
Y aun asi intentar separarla por 2 o 3 categorías, esto me obliga a repetir código

kenan-it-academy:
@orlandorojasoyarzun sounds reasonable!

orlandorojasoyarzun:
I think i misunderstood the message, I need to add more test and STILL try to separate the PR
@kenan-it-academy

orlandorojasoyarzun:
Hi @Kenan,

I went with the full battery of tests in one PR (6 tests covering 401, 403, 404, 422 x2, 200). I know this is bigger than we'd like, but splitting would make me duplicate code. Happy to split if you still prefer it, I already prepared two Sub Issues from this one, but according to the time schedule, is either I defend this PR, or the Issue will be fully blocked

The endpoint is fully working and tested. Ready for your review.

josemanuelriu-cmd (approved):
With the requested changes made, the PR is good to go as far as I'm concerned.

kenan-it-academy (approved):

PR #873, opened 2026-06-26
Feature/833 b notification endpoints

### LINKED ISSUE #834: Create endpoint to detect the league change notification
## Context
The frontend needs a single call on login to know whether the user's league changed recently.

After team feedback, we simplified the original approach: instead of tracking the "dismissed" state via a PATCH endpoint or relying on complex frontend localStorage cycles, the validation is entirely time-based and handled by the backend. The backend checks for a league transition within a 24-hour window, keeping the frontend completely stateless.

## Criteria
- GET /ligas/notification requires authentication (Sanctum).
- Queries the league_weekly_results table for the user's most recent entry.
- If a league change occurred in the last 24 hours:
Returns { hasChange: true, direction: 'up' | 'down', leagueName: string }
- If no recent change exists or the change is older than 24h:
Returns { hasChange: false }
- PHPUnit tests cover: no history, same league (no change), change older than 24h, promotion (up), and demotion (down).

## To Do
[x] Add getNotification() method to LigaController
[x] Add route GET /ligas/notification to api.php
[x] Add PHPUnit tests (asserting correct behavior, 24h window constraints, and JSON structures)

francobrida:
## Context
Adds the backend endpoint required to detect if the authenticated user had a league promotion or demotion in the last 24 hours.

## What this PR does
- `GET /ligas/notification` returns `{ hasChange: true, direction: 'up'|'down', leagueName: string }` if a league change occurred in the last 24 hours
- Returns `{ hasChange: false }` if no recent change exists or the change is older than 24h
- No PATCH endpoint — the frontend handles the dismissed state locally in React

## Why 24h?
Based on feedback from @kenan-it-academy: keeping the backend stateless by using the existing `league_weekly_results` table and a time window, rather than adding a dismissed flag.

## Files changed
- `app/Http/Controllers/LigaController.php` — added `getNotification()`
- `routes/api.php` — registered `GET /ligas/notification` as `ligas.notification`
- `tests/Feature/LigaTests/LigaNotificationTest.php` — 5 tests using `::create()` directly

## Acceptance criteria
- Returns `hasChange: false` when no history
- Returns `hasChange: false` when same league
- Returns `hasChange: false` when change is older than 24h
- Returns `hasChange: true` + `direction: up` for promotion
- Returns `hasChange: true` + `direction: down` for demotion

### DIFF
--- a/backend/src/app/Http/Controllers/LigaController.php
+++ b/backend/src/app/Http/Controllers/LigaController.php
@@ -126,37 +126,25 @@ class LigaController extends Controller
 
     public function getNotification(Request $request): JsonResponse
     {
-        $userId = $request->user()->id;
-
-        $latestResult = \App\Models\LeagueWeeklyResult::where('user_id', $userId)
-            ->orderBy('year', 'desc')
-            ->orderBy('week_number', 'desc')
+        $recentChange = \App\Models\LeagueWeeklyResult::where('user_id', $request->user()->id)
+            ->where('created_at', '>=', now()->subDay())
+            ->whereColumn('from_league', '!=', 'to_league')
+            ->latest()
             ->first();
 
-        if (!$latestResult) {
+        if (!$recentChange) {
             return response()->json(['hasChange' => false]);
         }
 
-        $fromLeagueId = $latestResult->from_league;
-        $toLeagueId = $latestResult->to_league;
-
-        if ($fromLeagueId === $toLeagueId) {
-            return response()->json(['hasChange' => false]);
-        }
-
-        $direction = $toLeagueId > $fromLeagueId ? 'up' : 'down';
-        
-        // Convert ID to name using the Enum if available, otherwise just mapping
-        $leagueNames = [1 => 'Bronze', 2 => 'Silver', 3 => 'Gold', 4 => 'Platinum'];
-        $leagueName = $leagueNames[$toLeagueId] ?? 'Desconeguda';
+        /** @var \App\Enums\LeagueTypeEnum $fromLeague */
+        $fromLeague = $recentChange->from_league;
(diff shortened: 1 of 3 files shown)


DanielVeraMunoz (approved):
Checked LeagueTypeEnum — it's int-backed (Bronze=1, Silver=2, Gold=3), so the up/down direction comparison is solid. Clean implementation, good test coverage for all 5 ACs. Approved!

orlandorojasoyarzun (changes requested):
Good work !, but I found a few things:

 - **Import statement missing** - in the method getNotification in `LigaController.php` uses `\App\Models\LeagueWeeklyResult` with full namespace. You could add the import at the top and use the short name.

`use App\Models\LeagueWeeklyResult;`

And then in the method getNotification:
// From
`$recentChange = \App\Models\LeagueWeeklyResult::where(...)`
// To
`$recentChange = LeagueWeeklyResult::where(...)`

 - **Missing null check for authenticated user** - While the route is protected by auth:sanctum, the method should defensively check if `$request->user()` is null and return 401 if so.

 - **Query optimization** - The `whereColumn` filter should come before the date filter to reduce the number of rows scanned first. Order matters for SQL performance ,)

francobrida:
Hi @orlandorojasoyarzun !

Done! Just pushed the changes covering all your feedback:
- Added the missing import for `LeagueWeeklyResult` at the top and cleaned up the controller namespace.
- Added a defensive null check for `$request->user()` returning a `401` if unauthenticated.
- Reordered the query filters to place `whereColumn` before the date check for better SQL index scan performance.
Tested it locally and all tests are passing green. 

Thanks for the review!

orlandorojasoyarzun (approved):

kenan-it-academy (approved):

PR #874, opened 2026-06-26
Feature/show all suggestions onstudent view front end

clara-cdp:
### what it does
It allows a student to toggle between viewing their own created tickets and viewing everyone's suggestions, regardless of user ID or role, when "veure suggestions" is clicked.

### what's changed
- ticketingPage (and test page)
- useTicketingGetAll hook (and testing)

> note: for local testing you will need to copy ticketController index() from  https://github.com/IT-Academy-BCN/ita-wiki-monorepo/pull/875
> note: button rendering on page student ONLY related PR: https://github.com/IT-Academy-BCN/ita-wiki-monorepo/pull/867

## Student view of own created Tickets

Screenshot 2026-06-26 082526

## Student view of all Suggestion Tickets
Screenshot 2026-06-26 082545

PR #875, opened 2026-06-26
Feature/show all suggestions onstudent view backend

### LINKED ISSUE #863: Add filter functionality to show suggestions button
As a student I will click the suggestions button and check out some else's suggestions.

> in the future I will be able to vote for my favourites

clara-cdp:
# Added logic to ticketController to allow students to view other users suggestions
### description
Modified the ticket controller's index() method to allow students to view—but not edit—everyone's tickets within the "suggeriments" category, even if they do not own them.

#### ticket controller: 
- imported ticketCategory and Request.
- index(): added logic that allows student view suggestions
TicketControllerTest has been also updated with a new test 

### out of scope
front end implementation 

> To understand better what this PR is related to: this UI frontend countepart  https://github.com/IT-Academy-BCN/ita-wiki-monorepo/pull/874


gregoridelrio (approved):
Clean implementation and good test coverage! Approved!

francobrida (changes requested):
Hi @clara-cdp! Nice work! The changes are very clean and the test updates are spot on.

Just one thing to check before merging: 

While `index()` now lets students list everyone's suggestions, the `show()` method in `TicketController.php` still returns a `403` if a student tries to open another user's suggestion by ID (since it only checks for creator/assignee/admin roles). 

Could you tweak `show()` to also allow viewing if the ticket category is a `Suggestion`, and add a quick test for it?

Also, quick question: does the frontend really need to toggle `include_suggestions` via query param, or should we just include suggestions by default for all students? this is just to understand the whole thing

Let me know what you think!

clara-cdp:
Hi Franco, 
Thank you for the review!
I didn't update the show method as, for now, there is no reason why a student would view only one suggestion. Should a student be able to see details of an specific suggestion? You're right to point this out (and I think that would be a correct next step, -one we didn't even take into consideration- and i guess that makes out of scope of this PR.
I hope I am answering your question: the API was set up to not to allow a student to view any ticket that wasn't owned. Also when a student is authenticated, it can only view their own tickets, not the UI cluttered with the rest. That is why I created a query parameter for a toggle button, as only suggestions (not other ticket categories) will be visible to students, and only upon request. The aim of it is that an upvoting system will be implemented, probably on the next sprint.

I hope this answer your doubts, if you think I still have to correct anything else, do let me know. 

thanks1

francobrida (approved):
Thanks for the clarification! 
Your explanation about the query parameter makes sense, nice planning!
guess the `show` method updates will be tackled maybe on the next sprint 

Approving the PR now. Great job!

clara-cdp:
Brilliant! thanks!

kenan-it-academy (approved):

PR #876, opened 2026-06-26
feat: implement league promotion logic

### LINKED ISSUE #831: Implement promotion and demotion logic
DESCRIPTION

Promote the top 3 students and demote the bottom 3 students between league tiers, so that the league reflects weekly performance.

OUT OF SCOPE

- Scheduling or triggering the process
- Resetting weekly points
- Frontend changes

ACCEPTANCE CRITERIA

- Top 3 students in Bronze are promoted to Silver
- Top 3 students in Silver are promoted to Gold
- Bottom 3 students in Silver are demoted to Bronze
- Bottom 3 students in Gold are demoted to Silver
- Students in Gold are never promoted further
- Students in Bronze are never demoted further
- Historical cumulative points (points) are not affected
- Logic is implemented inside ProcessLeaguePromotions command
- Test

gregoridelrio:
## Description
Implements promotion logic inside ProcessLeaguePromotions command.

## Done
- Top 3 students in Bronze are promoted to Silver
- Top 3 students in Silver are promoted to Gold
- Promotions only trigger if the league has at least 4 users (to avoid promoting everyone in a near-empty league)
- Idempotent: if promotions have already been processed for the current week, the command exits early without making any changes

## Out of scope (deferred to next PR)
- Bottom 3 students in Silver are demoted to Bronze
- Bottom 3 students in Gold are demoted to Silver

## Known issue
Frontend is displaying league tiers in reverse order.
A user promoted from Bronze to Silver appears visually as if they are in Bronze.
This is a frontend bug, out of scope for this PR.

## Tested
- Manual testing confirms promotions are applied correctly in the database


francobrida (changes requested):
Hey @gregoridelrio! 👋 Nice work on this one!

The logic is clean and well thought out. Nicely done with the DB::transaction wrapping both the history insert and the league update. Also, grabbing the user IDs in memory before updating prevents any unintended "chain promotions" in a single run. Great thinking there.

Tested it locally and everything runs fine. Just two things I wanted to flag before we merge:

### Scheduler is missing liga:process-promotions
In console.php line 20 it's only scheduled the reset:

```php
Schedule::command('liga:reset-weekly')->weeklyOn(0, '00:00');
```
But liga:process-promotions is not scheduled anywhere. So in production, every Sunday at midnight the scheduler will reset everyone's weekly points to 0 — but promotions will never fire automatically. right?

The order also matters here: promotions need to run before the reset, otherwise we'd be promoting users with 0 points. Something like this should do it:

```php
Schedule::command('liga:process-promotions')
    ->weeklyOn(0, '00:00')
    ->then(function () {
        Artisan::call('liga:reset-weekly');
    });
```
Minor thing — possible redundant registration:

Not 100% sure on this one, but I read that in Laravel 11, any class under app/Console/Commands that has a $signature gets auto-registered by the framework.

If that's the case, the closure you registered in console.php (lines 16–18) might be duplicating what the ProcessLeaguePromotions class already does on its own. Worth double checking — if it is redundant, we can just remove those lines and keep things tidy.

Everything else looks solid. Fix the scheduler thing and it's good to go

gregoridelrio:
Hey! Thanks for the review @francobrida , really helpful catches.

Fixed both things:

- Added liga:process-promotions to the scheduler, chained before liga:reset-weekly so the order is correct

- Removed the duplicate closure registration from console.php

One thing worth noting: Artisan::call('liga:process-promotions') was throwing CommandNotFoundException at runtime (both in tests and via the HTTP endpoint), so I ended up calling the command class directly via app(ProcessLeaguePromotions::class)->handle() in the controller. Same fix applied in the test setUp() methods.

Everything passing now.

francobrida (approved):
Hey Lauren! Everything looks good. The workaround of invoking the command directly via the container (app(ProcessLeaguePromotions::class)->handle()) works perfectly and all tests are passing.
Approving. well done!

clara-cdp (changes requested):
Hi Lauren,
Everthing looks well-structured and the PR is very focused. 
A minor issue:

If the endpoint is called twice in the same week, the promotion command exits early safely (idempotent). However, the controller continues and wipes out everyone's weekly points anyway...

you could either move the reset into the same weekly transition flow so it only runs when processing actually occurs, or have the command return whether it processed anything and only reset when appropriate.

That said, everything else LGTM

gregoridelrio:
Fixed! The weekly points reset is now conditional — it only runs if ProcessLeaguePromotions returns a successful exit code (0), so calling the endpoint twice in the same week won't wipe out everyone's points if no processing actually occurred.

clara-cdp (approved):
Hi Lauren, 

This looks much better: Checking LeagueWeeklyResult::where(...)->exists() as a guard clause perfectly handles early exit requirements without unnecessary database load.
LGTM

kenan-it-academy (approved):

PR #883, opened 2026-06-28
Feature/bug bounty UI

lxcxsito:
This PR improves and centralizes the ticket API endpoints, adding dynamic handlers for status and priority updates to ensure consistent routing. It also refactors the ticket update hook to properly separate priority and status update logic, and updates the corresponding tests to match these changes.


orlandorojasoyarzun (changes requested):
Hi Lucas! 

I found an issue that could be address:
1. **TypeScript error in config.ts** - The `EndPoints` type is a union of strings only. If you add `status: (id: number) => ...` and `priority: (id: number) => ...` to `END_POINTS.tickets`, TypeScript will fail to compile.

**fix:** I would suggest to not add functions to `END_POINTS`. Instead, build the URL directly in `updateTicketStatus`:
```typescript
const url = `${API_URL}tickets/${id}/status`;
This keeps the EndPoints type as strings (consistent with the rest 
of the project) and avoids the type error.

orlandorojasoyarzun:
@lxcxsito nice !

orlandorojasoyarzun (approved):

orlandorojasoyarzun:
@lxcxsito out of curiosity ... the Prettier commit is confusing; it's not actually applying Prettier, but it doesn't affect functionality.

josemanuelriu-cmd (changes requested):
Hi Lucas,
Just one thing regarding the code: there are no tests for `updateTicket` and `updateTicketStatus` in `endPointTickets.test.ts`.
If you can add those tests, I'll approve the PR.

On another note, I verified locally that updating the status and priority seems to work, but the values ​​revert to their previous state upon refreshing the screen. This appears to be a bug from a previous sprint, so it falls outside the current scope.
Regarding your PR, I did confirm that setting the project status to "closed" adds points to the owner in the league. For future sprints, we should specify that a ticket's status cannot be changed once it is closed; otherwise, one could repeatedly reopen and close the same ticket, allowing the user to gain points every time it is closed.

lxcxsito:
> Hi Lucas, Just one thing regarding the code: there are no tests for `updateTicket` and `updateTicketStatus` in `endPointTickets.test.ts`. If you can add those tests, I'll approve the PR.
> 
> On another note, I verified locally that updating the status and priority seems to work, but the values ​​revert to their previous state upon refreshing the screen. This appears to be a bug from a previous sprint, so it falls outside the current scope. Regarding your PR, I did confirm that setting the project status to "closed" adds points to the owner in the league. For future sprints, we should specify that a ticket's status cannot be changed once it is closed; otherwise, one could repeatedly reopen and close the same ticket, allowing the user to gain points every time it is closed.

I've added a test for updateTicketStatus, which is the functionality introduced in this PR. updateTicket wasn't modified as part of these changes.

josemanuelriu-cmd (approved):
Hi Lucas,
With the explanation and the added test, everything looks good to me.

Well done.

kenan-it-academy (approved):

